PR #3, opened 2023-01-23
chore: add config for prettier, tsconfig, eslint

cemms1:
## What does this change?

Adds the following to the base repo:
- editor config
- eslint config (taking defaults from guardian packages)
- ts config (taking defaults from guardian packages)
- prettier config (taking defaults from guardian packages)
- generic package.json scripts to make it easier to execute basic workspace-wide commands
- husky pre-push hook to stop us wasting time waiting for CI if we have basic linting/building failures

Additionally fixes some eslint warnings and errors in a very patchy way, but this is just a "hello world" app at the moment so we can address these in detail when we come to building features


dblatcher:
If we think it would be beneficial, we could do anther PR to configure eslint to enforce the prettier styling rules, link in the DCR repo does.
There's some docs here with a link to the recommended plugins, but also the warnings about why this isn't necessarily the ideal developer experience: https://prettier.io/docs/en/integrating-with-linters.html#docsNav
If we do go down this route, I'd say we should at least mark the prettier rules as 'warn' rather than 'error'.

dblatcher:
> Assuming you mean to install eslint-config-prettier:

I meant something slightly different -  `eslint-config-prettier` (which we have) says "eslint should ignore rules in the eslint config that conflict with the prettier rules". I was thinking of `eslint-plugin-prettier` which says "eslint should enforce the prettier rules".

Currently, if you (for example) change the tabs indents to spaces in one of the .tsx files, this breaks the "@guardian/prettier" rule "useTabs: true" but eslint doesn't report this as an issue.  (autoformat in VScode will still correct it for you though).

I don't have strong opinions on whether the stylistic rules should be enforced or not. The do on DCR and it can be quite annoying at times.

PR #5, opened 2023-01-24
allow path rewrites in proxy config for UI app

cemms1:
## What does this change?

I was reviewing PRs and frustrated at how we had to have `/api` prefixing our API endpoints in `newsletters-api` in order for them to be proxied to `/api` in `newsletters-ui` so had a dig for ways to do that.

What I wanted to do is to rewrite the proxy path so that we have 
- `/health` being the endpoint in `newsletters-api`
- `/api/health` being the endpoint from within `newsletters-ui` to use the `newsletters-api` endpoint above

Found this way to do it inside the vite config instead of in a separate proxy config file
https://vitejs.dev/config/server-options.html#server-proxy

Just a suggestion though, if we can find a way to nicely rewrite the path using the separate proxy file I'm happy to go that way


dblatcher (approved):
Sounds right to me!

idrise (commented):
Lush

PR #6, opened 2023-01-24
Display buttons in UI

codingWithRach:
## What does this change?

Uses Source button components to creates a basic structure for buttons available from the UI (this is not supposed to represent the final selection or appearance)

When clicking on one of the buttons, an alert is displayed that the option has not yet been implemented

## How to test

Run `yarn dev` from your IDE and view http://localhost:4200/

## Images

Screenshot 2023-01-24 at 15 17 25

Screenshot 2023-01-24 at 15 17 38


codingWithRach:
> Just wondering, is there a reason we're implementing Button from scratch rather than using the [Source Button component](https://github.com/guardian/csnx/tree/main/libs/%40guardian/source-react-components/src/button)?

Good point - I've replaced LargeButton with the pre-defined button from Source

idrise (commented):
Looking good !

PR #14, opened 2023-01-31
Dynamic forms from schemas

dblatcher:
## What does this change?

New re-usable Components:
- re-useable `SchemaForm` component to generate a set of controlled form fields, which takes a zod object schema and a 'data' value (of the type described by the schema) as the main props. Only string, string Enum, boolean and number fields are supported, but nested objects in the schema can be rendered using another `SchemaForm` (as is done for `emailEmbed` and `illustration`.
- The `SimpleForm` component - uses the `SchemaForm` to render a complete form, taking a schema, initial data and submit function props - IE provides a working user input for any simple data schema.
- `NewsletterForm` using the zodSchema for the `Newsletter` type to (mostly) dynamically generate the fields and generate validation warnings. The 'submit' button will just display a 'not implemented' alert. (this form is unlikely to be user-facing - it may be of use for an "admin" or "super-user" section of the UI that allows all fields to be editted)

New routes to test:
 - "/create/" :  the`NewsletterForm`. The 'submit' button will just display a 'not implemented' alert.
 - "/test-forms/: two example instances of the `SimpleForm`

Adds a function the the data-client project for deriving some string values from the news letter name.

## How to test

Run locally and click the 'create new newsletter' button to open the form page.
Editing the `newsletterSchema` with new primitive fields in the data-client project will change the fields rendered by the form.
Fields can be excluded from the form or marked as readOnly by changing the props on the `SchemaForm`s.

This should provide a flexible pattern for generating input forms and validating user input on the frontend, without the need to manually build forms for every data type. E.G. we could use the same pattern for a 'create thrasher' and 'create newsletter template' forms using the schema for the data types.

Adds a fair degree of complexity - the API for the Schema Form is flexible, but might not be that easy to use.
The `SimpleForm` is intended to address this by providing an easy to use component for generating an user interface that allows them to submit validated, type-safe input back to the application.

## Images
Screenshot 2023-02-14 at 15 32 26

## Accessibility
-   [x] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [x] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)


apps/newsletters-ui/src/app/components/NewsletterForm.tsx
```diff
@@ -0,0 +1,228 @@
...
+			)}
+
+			<button onClick={handleSubmit} disabled={hasWarnings}>
+				Create Newsletter
+			</button>
+
+			<NewsletterDetail newsletter={newsletter} />
+		</div>
+	);
+};
```
codingWithRach, line 228: This is a really long file - could some parts of it be split out to make it more manageable?

codingWithRach (commented):
This is great work!  A lot of effort up-front, but should make the code so much more self-sufficient further down the line

apps/newsletters-ui/src/app/components/NewsletterForm.tsx
dblatcher: I could split out the definition of the blank newsletter. 
Unfortunately, factoring code out of Functional Components is difficult as it tends to rely on hooks like `useState` which need to be defined within the function body. We probably could, try refactoring some of the code into custom hooks or higher-level functions, but I'm not sure how to do that without adding more complexity.
Its a good thought, though - It might be something to re-visit when we have some more implementations of the `SchemaForm` so we can see what abstractions or patterns would fit.

dblatcher:
I've made some updates to the description for this as it's probably the generic Form component that will be more useful that the `NewsletterForm`

codingWithRach (approved):

PR #15, opened 2023-02-02
Add sort, filter and hide/show features to the react table

codingWithRach:
## What does this change?

Adds features to the react table (currently used to display the list of all newsletters, although the table itself is generic) to allow:
- the data to be sorted in ascending/descending order for one of the columns
- the data rows to be filtered to only those containing the entered search term (this filters across all the data for a newsletter, not just the visible data)
- individual columns to be hidden. The intention is to provide the user with a pre-defined subset of columns (currently only 3 but this will be expanded in a separate PR), allowing them to hide the ones that are not of interest on a particular occasion

Styling will be addressed in a separate PR

## How to test

Run `npm run dev` in your IDE
View http://localhost:4200/newsletters/ in your browser

## Images

Screenshot 2023-02-02 at 16 28 29

## Accessibility


apps/newsletters-ui/src/app/components/NewslettersTable.tsx
```diff
@@ -3,14 +3,14 @@ import { Link } from 'react-router-dom';
 import type { Column } from 'react-table';
 import type { Newsletter } from '@newsletters-nx/newsletters-data-client';
 import { formatCellBoolean } from './Cell';
-import { Table } from './Table';
+import { Table } from './Table/Table';
```
dblatcher, line 6: This is just a stylistic thing so not important, but if you have a component defined in a folder, you can call the main component file `index.tsx` so you can import it from './Table' rather than './Table/Table'. It works the same way, but it can help keep the component imports statements on the same level.

dblatcher (approved):

PR #18, opened 2023-02-03
ci: add basic EC2 app for deployment

cemms1:
## What does this change?

- Adds capability to bundle `newsletters-api` and runs this in the CI workflow
- Switches references of `DEV` to `CODE` in the CDK apps to deploy
- Creates an EC2 app with basic configuration to run a bundled `index.cjs` file - expected to run off the base AMI created in Amigo (we created a Node18 AMI called `newsletters-node`)
- Configures a new Gu Cname for EC2 app with domainName supplied as a prop to the cloudformation app
- Adds an autoscaling deploy to the riff-raff.yaml config and added AMI references to the cloudformation deployment
- Set the host to `0.0.0.0` as Fastify defaults to running on `127.0.0.1` only. This is important for deployment in a container, since without this it can only be accessed from within the container. See the [fastify docs](https://www.fastify.io/docs/latest/Reference/Server/#listen)

## Still to do

Write a sort of "facade" for this to allow easier EC2 deployment in future

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -2,11 +2,9 @@
 . "$(dirname "$0")/_/husky.sh"
 
 # Check workspace before pushing
-npx nx format:check
-npx nx affected --target=lint,test,build --parallel=3
+npx nx affected --target=format:check,lint,test,build --parallel=3
 
 # Check CDK setup before pushing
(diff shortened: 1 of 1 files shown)


dblatcher (approved):
🥇

PR #27, opened 2023-02-20
Rachel/wizard state machine

cemms1:
## What does this change?

Basic implementation of a wizard to create a new newsletter, allowing users to move from one page to another, with basic validation of the fields

Some known issues:
- the data is not currently held between the pages e.g. if pressing the Back button, the data previously entered on the previous page is not populated (this will be addressed in a future PR)
- where the pages contain screenshots, these are slightly blurry.  This will be improved once we have regained access to the URL Image Signer

## How to test

Run `npm run dev` in your IDE 
http://localhost:4200/wizardtest/ 

## Images

Wizard page 1


dblatcher (approved):
As you said, there is more we need to do on this, but I think it makes sense to merge at this point and take it from there.

PR #29, opened 2023-02-21
Subbranch wizard step machine

dblatcher:
## What does this change?
 updates to wire the storage instance into the executeStep functions
the 'createNewsletter' step will add new entries into the storage, but we don't have a mechanism to pass the listId back to the frontend yet

also adds another schema for the 'pillar' step


codingWithRach (approved):
Looking fab - thanks!

PR #30, opened 2023-02-22
Retain data entered into wizard

codingWithRach:
## What does this change?

Shares the collected form data between the different pages of the wizard, so that it's repopulated after moving between Back and Next

Ensures that the data is not deleted if there's a validation error on the first page of the wizard

Some refactoring

NOTE: the code currently contains a number of console.log / console.table, which are useful during development but will be removed in a future PR

## How to test

Run `npm run dev` in your IDE
http://localhost:4200/wizardtest/

## Images

https://user-images.githubusercontent.com/74301289/220969349-c7d3eb79-db96-43ab-a243-1def927a51e7.mov


dblatcher (approved):
👍

PR #31, opened 2023-02-23
Dblacher/material UI

dblatcher:
## What does this change?

Uses material UI for the form elements used in the schema forms and to style the frame for the `SimpleForm` (used on the demo page only)

## How to test

run locally, try the 'test forms' and 'wizard test' pages from the nav

Nicer UX for data entry, with the built-in features on the MUI elements.

## Images

Screenshot 2023-02-23 at 15 38 05


codingWithRach (approved):
Looks fab

PR #33, opened 2023-02-24
Dblatcher/drafts view

dblatcher:
## What does this change?
Adds API routes, route loader and Components to display the Draft newsletters created in the wizard.

## How to test

Create a Draft using the wizard, use the "view draft newsletters" nav link to see it on the list.

## Images

Screenshot 2023-02-24 at 13 01 30
Screenshot 2023-02-24 at 13 01 11


codingWithRach (approved):
This is great.  Obviously we might want to display different/more fields further down the line, but this can easily be adapted later (like we still need to do for the list of current newsletters).  

If we allow users to fill out the wizard a bit at a time, rather than making them do it all at once, we may not make the fields mandatory on the individual pages - but we shouldn't let them proceed (whatever that turns out to mean) beyond the Draft stage until we have all the required information.  So perhaps a useful additional field to consider here would be a checkbox to show whether or not all fields that are needed in order to proceed beyond Draft have been entered?

PR #34, opened 2023-02-24
Dblatcher/dynamic wizard markdown

dblatcher:
## What does this change?

Wizard steps have an optional `dynamicMarkdown` method, which is passed the incoming (from the client) `WizardFormData` and outgoing (to the client) `WizardFormData` to produce a string of markdown to display to the user.

The existing `markDownToDisplay` property is renamed `staticMarkdown`.

If a step does not have a `dynamicMarkdown` method, the "currentStep" handler falls back to the `staticMarkdown`.

## How to test

Use the wizard - there are dynamic uses of the newsletter name on the 'Pillar" and "finished" steps

The responses to the user can be customised based on the data they sent and /or the resulting change on the server.

Adds some complexity - have tried to add some helpers to make defining and populating a template from the data a bit easier.

## images
Screenshot 2023-02-24 at 15 44 26

### DIFF
--- a/apps/newsletters-api/src/app/routes/currentStep.ts
+++ b/apps/newsletters-api/src/app/routes/currentStep.ts
@@ -90,8 +90,14 @@ export function registerCurrentStepRoute(app: FastifyInstance) {
 					.send({ message: 'No next step found', body: body });
 			}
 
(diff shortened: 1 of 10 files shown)


codingWithRach (approved):
Brilliant - will really improve the user experience in the wizard

PR #36, opened 2023-02-24
Display fields automatically calculated from the entered Name

codingWithRach:
## What does this change?

After the user enters a Name and presses the Next button, three new pages of the wizard are displayed showing the fields automatically calculated from the Name field, and allowing them to be modified if required.

Validation needs to be added to ensure that the entered fields do not already exist (in any other draft or actual newsletter)

## How to test

Run `npm run dev` and navigate through the wizard

## Images

https://user-images.githubusercontent.com/74301289/221370633-512dc1ce-1c6a-445d-bbac-019cc40985af.mov

### DIFF
--- a/libs/newsletter-workflow/src/lib/calculateFieldsFromName.ts
+++ b/libs/newsletter-workflow/src/lib/calculateFieldsFromName.ts
@@ -1,0 +1,34 @@
+import { deriveNewsletterFieldsFromName } from '@newsletters-nx/newsletters-data-client';
+import type { WizardFormData } from '@newsletters-nx/state-machine';
+
+export const calculateFieldsFromName = (
+	formData: WizardFormData,
+): WizardFormData => {
+	const derivedFields = deriveNewsletterFieldsFromName(formData.name as string);
+	const updatedFormData: WizardFormData = {
+		...formData,
+		...{
+			identityName: derivedFields.identityName,
+		},
+		...{
+			brazeSubscribeEventNamePrefix:
+				derivedFields.brazeSubscribeEventNamePrefix,
+		},
+		...{
+			brazeNewsletterName: derivedFields.brazeNewsletterName,
+		},
+		...{
+			brazeSubscribeAttributeName: derivedFields.brazeSubscribeAttributeName,
+		},
+		...{
+			brazeSubscribeAttributeNameAlternate:
+				derivedFields.brazeSubscribeAttributeNameAlternate
+					? derivedFields.brazeSubscribeAttributeNameAlternate[0]
+					: undefined,
+		},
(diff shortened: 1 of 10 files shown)


libs/newsletter-workflow/src/lib/calculateFieldsFromName.ts
```diff
@@ -0,0 +1,34 @@
+import { deriveNewsletterFieldsFromName } from '@newsletters-nx/newsletters-data-client';
+import type { WizardFormData } from '@newsletters-nx/state-machine';
+
+export const calculateFieldsFromName = (
+	formData: WizardFormData,
+): WizardFormData => {
+	const derivedFields = deriveNewsletterFieldsFromName(formData.name as string);
```
dblatcher, line 7: Great stuff! I would suggest refactoring this function slightly so :
- it takes the name as argument rather than casting it  from the `WizardFormData.name` - it would be a slightly more robust function that way ;and 
- it returns an object (`DerivedDraftFields`?) typed to only contain the derived fields instead of a new `WizardFormData` with the existing inputs include. Just to make the flow a bit clearer. We should still be able to spread that object into the `DraftWithoutId`  - I'll put another comment below on the other file.

maybe something like:

```typescript
export const calculateFieldsFromName = (
	formName: string,
): DerivedDraftFields => {
	const derivedFields = deriveNewsletterFieldsFromName(formData.name as string);
```

libs/newsletter-workflow/src/lib/executeCreate.ts
```diff
@@ -24,8 +25,10 @@ export const executeCreate: AsyncExecution = async (
 
 	const listId = stepData.formData ? stepData.formData['listId'] : undefined;
 	if (!listId) {
+		// TO DO - calculating fields from the Name at this point is not generic
+		// would it be better done somewhere else?
 		const storageResponse = await storageInstance.createDraftNewsletter({
-			...parseResult.data,
+			...calculateFieldsFromName(parseResult.data),
```
dblatcher, line 31: With the refactor of `calculateFieldsFromName` suggested above, we could change this part to be somethign like this:

```typescript
		// if the form has a non-empty string for a name, use it to calculate the derived fields
		const derivedFields =
			typeof parseResult.data.name === 'string' && !!parseResult.data.name
				? calculateFieldsFromName(parseResult.data.name)
				: {};

		const storageResponse = await storageInstance.createDraftNewsletter({
			...parseResult.data,
			...derivedFields,
			listId: undefined,
		});

```

dblatcher (approved):
Looking great! I made a couple of suggestions, but they could be for another PR if you wanted to merge as - is

codingWithRach:
Oh yes, that's much neater!

PR #40, opened 2023-03-01
State edit forms

dblatcher:
## What does this change?

- extracts the "render a form so the user can edit the formData in component state" logic from the `Wizard` component to `StateEditForm`
- support front end validation warnings on the forms 
- adds validation rules on the schemas for the `identityName` (kebab casing) and `brazeValues` steps (underscore casing)
- styling tweaks on the forms

## How to test

wizard still works - user is warned of improperly cased string values for the validated fields.

Easier to refactor and re-use the form.
Allows us to prevent input of 'incorrect' strings for values that need the right format.

Using the zod Schema to specify the required string formats is useful, but we should also add unit tests to make sure the rules are wording as intended.

## Images

https://user-images.githubusercontent.com/30567854/222093630-d7ee9abb-0c9b-4288-b83b-c68f76949ab3.mov


codingWithRach (approved):
Love it!

PR #44, opened 2023-03-02
Allows details of existing draft to be edited via wizard

codingWithRach:
## What does this change?

An Edit link has been added to the list of draft newsletters, allowing each to be edited using the Wizard.

A different first page (of the wizard) is displayed when editing an existing draft, as opposed to creating a new draft.  When clicking the Back button on the identityName step, it now returns to the Edit first page (regardless of whether Create or Edit was originally invoked), as this reflects the fact that the draft is assigned a listId and saved as soon as the user moves off the first step.

## How to test

Run `npm run dev`
Click the `View draft Newsletters` button
Click the `Edit` link next to one of the draft newsletters
The wizard will be pre-populated with data defined for the selected draft

## Images

Screenshot 2023-03-02 at 08 17 07


apps/newsletters-ui/src/app/components/DraftsTable.tsx
```diff
@@ -28,6 +31,13 @@ export const DraftsTable = ({ drafts }: Props) => {
 				accessor: 'theme',
 				sortType: 'basic',
 			},
+			{
+				Header: 'Wizard',
+				accessor: 'wizardListId',
+				Cell: ({ cell: { value } }) => (
+					<Link to={`/wizard/${value as string}`}>Edit</Link>
```
dblatcher, line 38: Not sure, but you might need to update this path after re-syncing with main, because of the routing changes

libs/newsletter-workflow/src/lib/steps/editDraftNewsletter.ts
```diff
@@ -0,0 +1,32 @@
...
+		},
+		next: {
+			buttonType: 'GREEN',
+			label: 'Next',
+			stepToMoveTo: 'identityName',
+			onBeforeStepChangeValidate: (stepData): string | undefined => {
+				const name = stepData.formData ? stepData.formData['name'] : undefined;
+				return name ? undefined : 'NO NAME PROVIDED';
+			},
+			executeStep: executeCreate,
```
dblatcher, line 29: maybe we should rename `executeCreate` if it is also used to start editing and existing draft

dblatcher (approved):
Great stuff - I think this would be best handled on another PR, but when you get to the end of the 'edit' wizard route, the markdown still says "{{name}} has been created".
We could fix that by providing more context to the `dynamicMarkdown` function - maybe giving it the whole `CurrentStepRouteRequest` instead of just the `WizardFormData` so it can check if there was an `id` on the request (which indicates its an 'edit' not a 'create')

libs/newsletter-workflow/src/lib/steps/editDraftNewsletter.ts
codingWithRach: It should be calling executeModify at this stage, as executeCreate is only needed for a new newsletter - thanks!

libs/state-machine/src/lib/setupInitialState.ts
```diff
@@ -0,0 +1,30 @@
...
+		throw new Error('no storageInstance');
+	}
+
+	const newsletterId = requestBody.id;
+	if (!newsletterId) {
+		return {
+			currentStepId: requestBody.stepId,
+		};
+	}
+	const newsletterIdAsNumber = +newsletterId;
```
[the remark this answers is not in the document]
codingWithRach: Neither did I - thank Idris for that one!

apps/newsletters-ui/src/app/routes/demo.tsx
```diff
@@ -15,13 +17,22 @@ export const demoRoute: RouteObject = {
...
 };
+
+function WizardLink(): ReactJSXElement {
+	const { listId } = useParams();
+	return (
+		<Container>
+			<Wizard id={listId} />
+		</Container>
+	);
+}
```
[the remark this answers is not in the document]
codingWithRach: This has been changed slightly in a subsequent PR in any case, so will be addressed there

PR #45, opened 2023-03-02
Allow UI to contain multiple wizards

codingWithRach:
## What does this change?

The Wizard was hardcoded so that it always displayed steps for gathering the info for a new newsletter, but the UI should allow multiple types of wizard to be invoked

A second wizard has been implemented, but more as a proof of concept than necessarily reflecting eventual implementation 

## How to test

Run `npm run dev`
The previously created wizard still functions when accessed via `Create Draft Newsletter` and via the `Edit` link on `View draft Newsletters`
The new wizard is accessed via the `Launch New Newsletter` button

We want to make the process as easy and clear as possible for the user, so need to consider how best to guide them through the process

## Images

Screenshot 2023-03-02 at 11 53 01


apps/newsletters-api/src/app/routes/currentStep.ts
```diff
@@ -47,27 +47,21 @@ export function registerCurrentStepRoute(app: FastifyInstance) {
...
-						  )
-						: setupInitialState();
+				if (body.buttonId !== undefined) {
+					result = await stateMachineButtonPressed(
+						body.buttonId,
+						{
+							currentStepId: body.stepId,
+							formData: body.formData,
+						},
+						newslettersWorkflowStepLayout,
```
dblatcher, line 59: As per your comment in the "newsletter-workflow" file, I think it would be clearer (and more scalable) is there were a separate `WizardLayout` object for the "launch newsletter" workflow. I think for that to work, the `CurrentStepRouteRequest` would need to include a "wizardId" field. The component now has the wizardId prop, so that should help.

libs/newsletter-workflow/src/lib/newsletter-workflow.ts
```diff
@@ -1,20 +1,26 @@
...
+import { editDraftNewsletterLayout } from './steps/editDraftNewsletterLayout';
 import { finishLayout } from './steps/finishLayout';
 import { identityNameLayout } from './steps/identityNameLayout';
+import { launchNewsletterLayout } from './steps/launchNewsletterLayout';
 import { ophanLayout } from './steps/ophanLayout';
 import { pillarLayout } from './steps/pillarLayout';
 
+// TODO - this list currently contains steps from *all* wizards
+// should separate wizards be separate out?
 export const newslettersWorkflowStepLayout: WizardLayout = {
```
dblatcher, line 15: I think the answer is yes, there should be a separate `WizardLayout` objects for the "launch" and "create" wizards. As noted above, we'd need the frontend to provide a wizardId so the route knows which objects to find the step in.
One benefit of this is that the stepIds won't need to be completely unique - EG we could have multiple layouts with steps called 'start' and 'finish'.

apps/newsletters-api/src/app/routes/currentStep.ts
codingWithRach: I have separated out the layouts

libs/newsletter-workflow/src/lib/newsletter-workflow.ts
codingWithRach: I have separated out the layouts

PR #59, opened 2023-03-09
Create pages for data collection wizard

codingWithRach:
## What does this change?

Pages have been added to the data collection wizard to represent all the data currently being collected in the [P&E newsletter information form](https://docs.google.com/document/d/1wjaEXk5rPqGG7RVi02kVS-wKfiTBhkmjFG9LzE9ND6w/edit)

The wizard is incomplete (see comments below), but it is suggested that additions are addressed in a separate PR in order to keep this one a manageable size.

## How to test

`npm run dev`
Select Newsletters from the ribbon
Click the Demo create newsletter wizard button

## Images

https://user-images.githubusercontent.com/74301289/225257590-d4fa8498-b364-41d5-b595-3b2d37d7f051.mov

## Accessibility

### DIFF
--- a/libs/newsletter-workflow/src/lib/steps/newsletterData/formSchemas.ts
+++ b/libs/newsletter-workflow/src/lib/steps/newsletterData/formSchemas.ts
@@ -117,9 +117,9 @@ export const formSchemas = {
 
 	tags: z
 		.object({
-			seriesTag: z.string().url().optional(),
-			composerTag: z.string(),
-			composerCampaignTag: z.string(),
+			seriesTag: z.string(),
+			composerTag: z.string().optional(),
+			composerCampaignTag: z.string().optional(),
 		})
 		.describe('Input the tag setup'),
 
--- a/libs/newsletter-workflow/src/lib/steps/newsletterData/tagsLayout.ts
+++ b/libs/newsletter-workflow/src/lib/steps/newsletterData/tagsLayout.ts
@@ -11,7 +11,7 @@ const markdownTemplate = `
 
 Please share the series tag URL for the newsletter.
 
-For example: [https://www.theguardian.com/tv-and-radio/series/what-s-on-tv](https://www.theguardian.com/tv-and-radio/series/what-s-on-tv) for the What's On newsletter.
+For example: [tv-and-radio/series/what-s-on-tv](https://www.theguardian.com/tv-and-radio/series/what-s-on-tv) for the What's On newsletter.
 
 *If the tag does not already exist, an email will automatically be sent to Central Production to request its production.*
 


codingWithRach:
Some comments following run-through with Sophie:
- Testing period can be added automatically (i.e. not optional) - though can we make it private if need be 
- Allow specification of email addresses to which to send test newsletters
- Add async newsletter in frequency e.g. weekly for eight weeks
- Specify the series tags requires email to Central Production to create tags (could potentially be automated in future?)
- Should a confirmation email be sent - is this needed or automatically done through Braze (set to false for all newsletters in PROD spreadsheet)?
- Sign up image page could render the page as a preview
- Thrasher - shall we add tickbox to say is this visible on app as well as web? This is currently a manual reply email to Central Production
- Sophie will add step screenshots into a Google folder
- Display Braze, Identity and Ophan values as default (non-editable) and then offer link to modify if need be
- Add an editable workflow as markdown?  @Idris Elliott is adding a ticket

libs/newsletter-workflow/src/lib/steps/newsletterData/brazeLayout.ts
```diff
@@ -1,17 +1,33 @@
...
-const markdownToDisplay = `
+const markdownTemplate = `
 # Modify Braze Values
 
 These are tracking fields used by Braze.
 
-They have been calculated automatically from the Name that you entered on a previous step of the wizard, but you can change them if you need.
+They have been calculated automatically from the name **{{name}}**, but you can change them if you need.
+
+***TO SIMPLY THE DATA COLLECTION PROCESS, THESE WILL BE DISPLAYED READ-ONLY WITH A LINK TO EDIT IF REQUIRED***
```
dblatcher, line 14: The best way to include the values read-only would be to include them in the mark-down. We don't need to link to another page to edit them, it could just be an optional step that include the schema to edit the values ie

```mermaid
flowchart LR
    B[step: Readonly] --> C{buttons}
    C -->|edit button| D[Step: Edit] -->E
    C -->|ok button| E[Step for next question]
```

libs/newsletter-workflow/src/lib/steps/newsletterData/emailCentralProductionLayout.ts
```diff
@@ -0,0 +1,43 @@
...
+export const emailCentralProductionLayout: WizardStepLayout = {
+	staticMarkdown,
+	dynamicMarkdown(requestData, responseData) {
+		if (!responseData) {
+			return staticMarkdown;
+		}
+		const [name = 'NAME'] = getStringValuesFromRecord(responseData, ['name']);
+		return markdownTemplate.replace(regExPatterns.name, name);
+	},
+	buttons: {},
```
dblatcher, line 42: should this step have any buttons?

libs/newsletter-workflow/src/lib/steps/newsletterData/finishLayout.ts
```diff
@@ -4,7 +4,11 @@ import { regExPatterns } from '../../regExPatterns';
 
 const markdownTemplate = `# Finished
 
-You have reached the end of the wizard. The draft newsletter **{{name}}** has been created.`;
+You have reached the end of the wizard.
+
+Congratulatoions - the newsletter **{{name}}** has been launched.
+
+`;
```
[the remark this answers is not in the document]
codingWithRach: As Idris suggested last week, there probably isn't any need for a separate wizard in this particular case - we can have a single wizard which allows collection of the data, and then proceeds beyond to start doing the automated stuff.  If we go down that route, I'll need to rename some of the variables to reflect that this is a single wizard again!

libs/newsletter-workflow/src/lib/steps/newsletterData/brazeLayout.ts
codingWithRach: Good idea

libs/newsletter-workflow/src/lib/steps/newsletterData/emailCentralProductionLayout.ts
codingWithRach: Deliberately doesn't have any buttons, has the user should not have any way of moving to a different step of the wizard.  The theory is that the newsletter will be stuck on this step until an appropriate response from CP has been detected, and the wizard will then automatically move to the next step (possibly with notification sent to someone?)

libs/newsletter-workflow/src/lib/steps/newsletterData/finishLayout.ts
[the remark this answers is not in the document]
codingWithRach: In any case, due to the lack of buttons on the Email Central Production page (and the fact that, as yet, no behind-the-scenes processing is actually taking place), there is currently no way of reaching this page

codingWithRach:
As yet this wizard is incomplete, but omissions will be addressed in future PRs:
- the schema has not yet been defined for the Date page (as this is being addressed in a separate PR), and so there are no fields displayed (and therefore no validation)
- screenshots have not been added/updated
- where multiple values can be collected for a field (e.g. multiple LinkList trigger subheadings) the step is currently only allowing entry of a single string. To support an array of strings, changes will need to be made to the Schema Form component and WizardFormData type. 
- consideration needs to be given to how to collect multiple information for the ReadMore step, given that there are three fields for each trigger.  Possibly: on the first ReadMore page have an additional button to allow addition of another (showing the same page again).  However, will also need to have a way of deleting entries.
- consideration needs to be given to how best to collect information about multi-thrashers (currently it's just a tick box indicating that one or more multi-thrashers are required)
- queries are displayed in CAPITAL BOLD ITALIC

There have been separate discussions about
- what wizard button would be appropriate on each step and what colour they should be (currently we have a red Back and a green Next, but this is likely to change)
- needing to remember which step a user was on, so that they don't always have to start right at the beginning when editing a newsletter but instead can be taken straight to the next field not yet completed (if still in the data collection portion of the wizard) or the current step (if they've progressed beyond data collection)
- the data collection and subsequent steps have now been combined into a single wizard - this therefore needs to be renamed to reflect the dual purpose

libs/newsletter-workflow/src/lib/steps/newsletterData/formSchemas.ts
```diff
@@ -40,9 +43,92 @@ export const formSchemas = {
...
+		})
+		.describe('Input the Read More setup'),
+
+	tags: z
+		.object({
+			seriesTag: z.string().url().optional(),
+			composerTag: z.string(),
+			composerCampaignTag: z.string(),
+		})
+		.describe('Input the tag setup'),
```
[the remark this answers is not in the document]
codingWithRach: I only made seriesTag optional because of the auto validation caused by having it as a url.  You're right that it isn't really a url (so I need to update the example text on that page of the wizard) - I will therefore remove the "optional" for seriesTag - we're not expecting it to exist, but need to know what it should be so that we can automatically email CP to request its creation if it doesn't

dblatcher (approved):

PR #60, opened 2023-03-13
Use new newsletterData type

dblatcher:
## What does this change?

The API uses `NewsletterData` as the model for launched (as opposed to draft) newsletters rather than `LegacyNewsletter`.
 - The API newsletter routes expect the data source to provide items in the `NewsletterData` format.
 - The /legacy route converts valid items to  `LegacyNewsletter` before returning
 - The generate-newsletters script uses the new format
 - The Newsletters table view now renders some of the 'new' fields (creationTimestamp, cancellationTimestamp, status)

Note that the current version of `NewsletterData` is not intended to be 'final' - it can still be modified in future PR's provided the transform function is still able to correctly convert to `LegacyNewsletter`.

## How to test

use `npm run generate-newsletters` to get a new set of test data.
`nom run dev` to start the UI and API - the new items will be loaded using the `NewsletterData` model. The legacy API route will render then converted to `LegacyNewsletters`

We can change and extend the `NewsletterData` type as required, provided we are still able to convert the new model back to a `LegacyNewsletter`. This will allow us to add new features which require extra properties in the data model.
The test coverage on the transform module should alert to breaking changes.

## Images
Screenshot 2023-03-13 at 13 23 43


apps/newsletters-api/open-api.yaml
```diff
@@ -184,7 +182,92 @@ paths:
 
 components:
   schemas:
-    # TO DO - define schema for Newsletter when implemented in code.
+    Newsletter:
```
codingWithRach, line 185: Note to self: once PR #59 has been completed, data collected through the wizard can be added to this schema

libs/newsletters-data-client/src/lib/newsletter-data-type.ts
```diff
@@ -1,22 +1,26 @@
...
+	nonEmptyString,
+	underscoreCasedString,
+} from './schema-helpers';
 
 /**
  * NOT FINAL - this schema a placeholder to test the data transformation structure.
  * Edits to this schema would need to be reflected in the transform function.
  * The actual data model is TBC.
  */
 export const newsletterDataSchema = z.object({
```
codingWithRach, line 15: Note to self: will also need to update here once PR #59 completed

tools/scripts/deno/generate-newsletters.js
```diff
@@ -1,5 +1,7 @@
 import { faker } from 'https://cdn.skypack.dev/@faker-js/faker@v7.6.0';
 
+const YEAR_IN_MS = 1000 * 60 * 60 * 24 * 365;
```
codingWithRach, line 3: Debatable whether this is a year in ms (leap years?) but for the purpose of generating newsletters it's close enough :-)

codingWithRach (approved):

PR #61, opened 2023-03-14
Dblatcher wizard step nav

dblatcher:
## What does this change?

Uses an MUI `<Stepper/>` component to show progress through the Wizard.

Adds optional 'label' and 'role' properties to the `WizardLayoutStep` to control how it appears in the stepper (eg the step used as the starting point for editing and item should not appear in the list of steps when the user is creating a new item, the 'cancel' step is not ever shown)

## How to test

test the wizards in development

Users get a visual representation of progress through the Wizard

## Images

Screenshot 2023-03-13 at 17 32 37


apps/newsletters-ui/src/app/components/StepNav.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+interface Props {
+	currentStepId?: string;
+	stepList: StepListing[];
+	onEditTrack: boolean;
+}
+
+export const StepNav = ({ currentStepId, stepList, onEditTrack }: Props) => {
+	const filteredStepList = stepList.filter((step) => {
+		switch (step.role) {
```
codingWithRach, line 12: In my IDE I'm getting "Unsafe. member access .role on an 'any' value" (ditto id and label on lines 27 and 28)

libs/state-machine/src/lib/getStepList.ts
```diff
@@ -0,0 +1,12 @@
+import type { WizardLayout, WizardStepLayout } from './types';
+
+export type StepListing = {
```
codingWithRach, line 3: If we also include the "step" here, can we allow the user to click on one of the breadcrumbs in order to move straight to that step?

codingWithRach (approved):
Looks brilliant.  As already discussed we might have to revisit the visual once the wizard(s) are complete and we have a huge number of steps, but that's no reason not to include the functionality now

PR #62, opened 2023-03-14
support date fields

dblatcher:
## What does this change?

Supports Date fields in the in schema forms.
Adds a step to the draft Wizard for specifying the desired dates for the sign-up pages and thrashers to go live.
uses the data values in the dynamic markdown for the 'finish' step

## How to test

run the wizard - dates values can be set and edited.

Date values can be collected in the wizard.

The values are typed as dates for validation, but converted to strings when serialised to JSON - the UI needs to handle the values in either format and coerce strings to dates (Zod supports this). 
Although this works, it adds complexity to the data flow and could be a source of future 

## Images
Screenshot 2023-03-14 at 12 06 46
bugs if not handled correctly.


libs/newsletter-workflow/src/lib/steps/newsletterData/promotionDateLayout.ts
codingWithRach, line 1: Now that PR #59 has been merged, this file is not needed (the schema can instead be added to the existing dateLayout.ts page)

libs/state-machine/src/lib/getEmptySchemaData.ts
```diff
@@ -23,6 +23,8 @@ export const getEmptySchemaData = (
 			mod[key] = 0;
 		} else if (zod instanceof ZodBoolean) {
 			mod[key] = false;
+		} else if (zod instanceof ZodDate) {
+			mod[key] = new Date();
```
codingWithRach, line 27: Is it possible to have a date field without defaulting it to today?  In my past experience, if a user is unsure what value something should have they often just accept the default, no matter how inappropriate.  If it's mandatory to have a default, then it might be better to be e.g. two week's time or a month's time, rather than today

dblatcher: 👍 Date fields can default to 'undefined' - this does mean that when the field is not optional on the schema, the form starts out with a validation error, but that's probably better than using an inappropriate default.

I think with a bit of work, we could define defaults on the schema for particular fields, but that's one for another PR!

codingWithRach (approved):

PR #63, opened 2023-03-14
Dblatcher/design link property

dblatcher:
## What does this change?
- adds the `designUrl` property to `NewsletterData` (must be a valid URL or undefined)
- add a step allowing the users to specify and optional link to a design page for the newsletter 
- include the link in the table view for Newsletters and Drafts

[depends on earlier PR: https://github.com/guardian/newsletters-nx/pull/60]

## How to test

Run the create newsletter Wizard. On the new step the form will validate that the value entered is a valid URL, but will accept an empty string (parsed as undefined)
If you add a value, the link button will be rendered on the draft table view.

## Images

Screenshot 2023-03-14 at 17 37 07


codingWithRach:
Some overlap between this and PR #59?  The latter has a wizard page requesting the Figma url (amongst other things)

libs/newsletter-workflow/src/lib/steps/newsletterData/descriptionLayout.ts
```diff
@@ -6,13 +6,11 @@ import { executeModify } from '../../executeModify';
 import { formSchemas } from './formSchemas';
 
 const markdownToDisplay = `
-# Specify the Description
+# Is there a design page?
```
codingWithRach, line 9: Don't think this bit is needed, in light of PR #59

libs/newsletter-workflow/src/lib/steps/newsletterData/designLinkLayout.ts
codingWithRach, line 1: Don't think this is needed, in light of PR #59

codingWithRach (commented):
Looks great, other than the changes to wizard pages (which are fine as a standalone version, but I don't think are needed in light of PR #59)

codingWithRach (approved):

PR #69, opened 2023-03-23
Use newsletter data schema in form schema

dblatcher:
## What does this change?

 - Adds all data points collected in the wizard to the `newsletterDataSchema` (from which the `NewsletterData`) type is derived.
 - Uses the `newsletterDataSchema` to define the schemas for each wizard step so the definitions are not duplicated and the data types must match.
 - Updates the newsletter data generator to include the new required properties of `NewsletterData`.
 - refactors the "transform to legacy newsletter" function - this now manually sets each field on `LegacyNewsletter` instead of merging the data types and destructuring, so it will not longer be necessary to amend the function to exclude every new field in `NewsletterData` that should not be included in `LegacyNewsletter`

## How to test

API will return lists of legacy newsletters and newsletters.
Wizard can still collect data as before, with all points being stored.

Will be easier to maintain consistency between our wizard steps and the data model.
Metadata(eg field descriptions, custom validation warning messages) set on the `newsletterDataSchema` will be inherited by the schemas used on the wizard step - will allow future PR's making use of these on the UI.


libs/newsletter-workflow/src/lib/steps/newsletterData/formSchemas.ts
```diff
@@ -1,134 +1,145 @@
...
+		'readMoreWording',
+		'readMoreUrl',
+	]).describe('Input the Read More setup'),
 
-	tags: z
-		.object({
-			seriesTag: z.string(),
-			composerTag: z.string().optional(),
-			composerCampaignTag: z.string().optional(),
+	tags: newsletterDataSchema
```
idrise, line 124: zod fu is occurring.

idrise (approved):

PR #71, opened 2023-03-27
Dblatcher/schema form array input

dblatcher:
## What does this change?

Adds a component for editing string arrays, and extends the `SchemaForm` component to render it when passed a schema with a `z.array(z.string())` field.

Adds extra tests to account the validation of members of a string array - eg for `z.array(underscoreCasedString())`. 

### ...with the steps as they are 
Start the "create draft" wizard. got through the steps to *13.Read More Sections*, *14.Link List Sections* and *18.Braze* to see the array input.

The array can be changed and will be validated against the schema and stored in the `draftStorage` like any other value.

### ...quickly
Update the libs/newsletter-workflow/src/lib/steps/newsletterData/formSchemas.ts so the layout after the initial "enter name" step includes a string array field (this is just to avoid going all the way to the Braze step when testing)  EG:

```typescript
	promotionDates: newsletterDataSchema
		.pick({
			signUpPageDate: true,
			thrasherDate: true,
			brazeSubscribeAttributeNameAlternate: true,
		})
		.describe('choose the dates you want promotions to appear'),
```

Can use the string array input when needed. Component should be easy for users to understand.

Seems to work safely - test coverage is in place.

### 1 valid item
Screenshot 2023-03-27 at 13 59 17

### validation warning
Screenshot 2023-03-27 at 13 59 50

### 2 valid items
Screenshot 2023-03-27 at 14 00 18

## empty list
Screenshot 2023-03-27 at 14 00 28


apps/newsletters-ui/src/app/components/ArrayInput.tsx
```diff
@@ -32,36 +42,63 @@ export const ArrayInput = ({
...
+								variant="outlined"
+								title={`delete entry "${item}"`}
+								onClick={() => {
+									handleDelete(index);
+								}}
+							>
+								x
+							</Button>
+						</Grid>
+					</Fragment>
```
idrise, line 80: React added <></> as a more modern update to fragments. Just another way to skin the same rabbit.

PR #74, opened 2023-03-30
Dblatcher/separate wizard paths

dblatcher:
## What does this change?

 - Adds a third layout ("RENDERING_OPTIONS") to the "newsletter-workflow" wizard.
 - Moves the step about the email rendering template to the new layout, reducing the number of  steps 'NEWSLETTER_DATA' steps from 22 to 16
 - Adds link buttons to the draft details and list table to access either the  RENDERING_OPTIONS or NEWSLETTER_DATA(edit) layouts for an existing draft.
 - add a utility function that uses the `WizardStepLayout.role` property (first added to help render the progress bar) to determine the initial step in any wizard, rather than manually defining for each layout.
- adds links to the draft details page in the markdown for the 'finish' steps.

## How to test

Create a new draft using the 'create' wizard. The option to use the RENDERING_OPTIONS wizard will be on the drafts table and draft details page

## Images

Screenshot 2023-03-30 at 17 32 05

Screenshot 2023-03-30 at 17 33 03

Screenshot 2023-03-30 at 17 33 46


codingWithRach:
I haven't looked at the code yet, but a couple of general points:
1. as the purpose of the UI is (eventually) to automate (as much as possible of) the creation of newsletters, I rather assumed that it would only be used for the creation of newsletters using email-rendering.  However, your comment "This wizard is to choose the options for how an article-based newsletter will appear in Email-rendering.  You do not need to complete this wizard for fronts-based newsletters or newsletters that do not use Email-rendering." suggests the UI will be used for other types of newsletter too.
2. I have suggested to Sophie that we get Editorial involved in order to determine what they would find more usable e.g. a single wizard with lots and lots of steps that guides you through in order, or lots of separate wizards so that the number of steps is not so overwhelming
3. One of the ideas mentioned previously was that there was no need to have a separate list of Draft newsletters, but that they could be included in the same list as the live newsletters (as long as it's clear which are live and which aren't, and some options e.g. Delete would only apply to Draft entries).  Before we add more functionality to the Drafts table, we probably need to decide whether this is going to happen (might be another usability question for Editorial)

It might just be because I've been away, but I'm feeling a lack of clarity about what the tool is going to be used for, and how it should work!

codingWithRach:
Also, not related to this PR, but now that data model has been changed, the field labels are even less user friendly.  Would it be possible to display a Label rather than the actual field name?
Screenshot 2023-04-05 at 08 25 31

codingWithRach:
And would these be better displayed in a separate table?  Again, not a very user-friendly format, especially with the need to scroll
Screenshot 2023-04-05 at 08 28 05

libs/newsletter-workflow/src/lib/steps/launchNewsletter/noItem.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const noItemLayout: WizardStepLayout = {
+	staticMarkdown: `# Launch a newsletter
+
+This wizard will guide you through the process of launching a new newsletter using email-rendering.
+
+**no draft has been selected to launch**
+
+`,
+	role: 'CREATE_START',
```
codingWithRach, line 11: Need to add a label so doesn't display as "noItem" in the breadcrumbs

codingWithRach (approved):

dblatcher:
All good questions!

> 1. as the purpose of the UI is (eventually) to automate (as much as possible of) the creation of newsletters, I rather assumed that it would only be used for the creation of newsletters using email-rendering.  However, your comment "This wizard is to choose the options for how an article-based newsletter will appear in Email-rendering.  You do not need to complete this wizard for fronts-based newsletters or newsletters that do not use Email-rendering." suggests the UI will be used for other types of newsletter too.

The first aim of the project overall was to replace the existing API and the spreadsheets, so it needs to support all the types of newsletters it contains (article based, fronts based, other/manual). The project will need to provide a way for users to create and edit any type of newsletter.

> 2. I have suggested to Sophie that we get Editorial involved in order to determine what they would find more usable e.g. a single wizard with lots and lots of steps that guides you through in order, or lots of separate wizards so that the number of steps is not so overwhelming

> 3. One of the ideas mentioned previously was that there was no need to have a separate list of Draft newsletters, but that they could be included in the same list as the live newsletters (as long as it's clear which are live and which aren't, and some options e.g. Delete would only apply to Draft entries).  Before we add more functionality to the Drafts table, we probably need to decide whether this is going to happen (might be another usability question for Editorial)

That's a good point - if we did decide to scrap or rework the Drafts table, the functionality for launching the 'rendering options' wizard is contained in the link button (which is also on the draft detail page) so we won't loose much if we decide not to have the button on on the table.

dblatcher:
> Also, not related to this PR, but now that data model has been changed, the field labels are even less user friendly. Would it be possible to display a Label rather than the actual field name? Screenshot 2023-04-05 at 08 25 31

Yes - I think the best way to do this would be to use the [describe](https://github.com/colinhacks/zod#describe) method in the schema to provide user-friendly text that the `SchemaForm` could use for the label on the form elements. It could fall back to the field key so we won't need to describe every field.

dblatcher:
> And would these be better displayed in a separate table? Again, not a very user-friendly format, especially with the need to scroll Screenshot 2023-04-05 at 08 28 05

Yes - we could have a custom display for each draft property when the the data model is finalised, but I didn't want the Components to block changes to the schema, so for now its just running through the keys and outputting values programatically.

I've made a small change for no so it at least formats the object JSON, but I'd still think of this table as a temporary measure until we know exactly what data we'll have to display.
Screenshot 2023-04-05 at 11 19 06

codingWithRach:
> > And would these be better displayed in a separate table? Again, not a very user-friendly format, especially with the need to scroll Screenshot 2023-04-05 at 08 28 05
> 
> Yes - we could have a custom display for each draft property when the the data model is finalised, but I didn't want the Components to block changes to the schema, so for now its just running through the keys and outputting values programatically.
> 
> I've made a small change for no so it at least formats the object JSON, but I'd still think of this table as a temporary measure until we know exactly what data we'll have to display. Screenshot 2023-04-05 at 11 19 06

That makes a lot of sense, but in the meantime the formatting makes it a lot more readable :-)

libs/newsletter-workflow/src/lib/steps/launchNewsletter/noItem.ts
dblatcher: 👍 I'll label it as 'start' and use the same label for the `launchNewsletterLayout` (the initial step for when there is a selected item)

PR #76, opened 2023-04-03
Dblatcher/s3 env vars

dblatcher:
## What does this change?

Removes the hard-coded s3 configuration, has the storage service read the s3 params from environment variables instead. 

The application will use an `S3DraftStorage` if all three required s3 params are present in the environment variables, unless the USE_LOCAL_STORAGE variable is true

gitignores .env.local files at any path.

## How to test

Locally:
 - Copy and rename **apps/newsletters-api/env.local.example.txt** to  **apps/newsletters-api/.env.local**
 - update the S3_BUCKET value in .env.local to then name of an s3 bucket in the "developerPlayground" account

run locally, `npm run dev` and open http://localhost:3000/api/drafts - if you don't have developerPlayground from [janus](https://janus.gutools.co.uk/), the request should fail with a 403 error. If you get credentials (and restart the application), it should list the draft files in the bucket (if any).

Setting USE_LOCAL_STORAGE=true in your **.env.local** (and restart the application). [The API page](http://localhost:3000/api/drafts) will show the 5x sample drafts hard coded in the `makeInMemoryStorageInstance` function.

Moves us closer to a "real world" implementation that could work in PROD. No hard coded S3 parameters.

We should still request help from DevX to adapt this for production - not sure how best to obtain s3 credentials on CODE and PROD (use parameters store / secrets manager?).


idrise (approved):
I think we are getting to the point of having a little complexity and we still have the default readme, sounds like one to add as a def of done. Rambling and not connected to the work only the work contains knowledge that we will need to share eventually.

PR #77, opened 2023-04-04
Dblatcher/links in step nav

dblatcher:
## What does this change?

The Stepper nav will render buttons allowing users to skip to the step if the the `WizardStepLayout` has the new `canSkipTo` property is true. The skip buttons will trigger a request to the state-machine which will:
 - validate the formData on the current step against the schema, and:
   - keep the user the same step and return a warning message if the data is not valid
   - update the storage and navigate to the request step if not.

Note that the custom hooks associated  the current step's 'next' button are not triggered by skipping using the stepper nav as no wizard button was pressed. 

**possible change - the `WizardStepLayout` could have another new optional property to specify which button's hook's to use when the user skips**

The data input steps (IE not start, finish, cancel) for the rendering options layout are set to `canSkipTo=true`

## How to test

Go to the drafts table and click the "edit rendering options" button for an existing draft. 

Users able to move quickly to a particular step

Allowing skipping on wizards that collect essential data or trigger external actions could lead to gaps - docs for writing layout should have guidance on when to allow skipping (only on wizards to edit sets of optional info?).

## Images

https://user-images.githubusercontent.com/30567854/229734418-20059278-a99d-47d5-8c8a-37f9025ce78e.mov

## Accessibility


codingWithRach:
There would be an argument for setting the property to true for all data collection steps in all wizards, as editorial often don't have all the information when they start filling out the existing info form, but instead skip around, so so it might (in real life) prove to be annoying to force them always to go through the wizard in order.  Having said that, perhaps the wizard will result in a different way of working, in which the wizard is not filled out until all the information is available?  

But even if that's the case when creating a new draft, we really should set the property to true for all data collection steps when editing an existing draft (until Data Collection is deemed complete, and the automated processes have started - when they exist), as it would be really annoying to have to go through the entire wizard just to edit something on the penultimate step.  At the point the user is claiming Data Collection is complete, we need to add validation to ensure that all required fields have been filled out.

As @dblatcher mentioned yesterday, there is an argument for removing most of the validation we've added to individual steps, allowing this to be handled by the schema.  Some validation hasn't yet been implemented (e.g. name/braze info doesn't already exist), and some validation links data between multiple steps (e.g. if you say you want thrashers, then you must confirm that the Figma design includes thrashers) - this validation might arguably be best left where it is for context (but perhaps repeated at the completion of Data Collection to ensure it hasn't been skipped and that nothing has changed in the meantime)

codingWithRach:
Some consideration needs to be given to what should be displayed on the Finish page for the Rendering Options wizard.  Currently it says:
Screenshot 2023-04-06 at 07 32 22

However, now that the user can skip around, there may be some pages that haven't been filled out, in which case it's not really accurate to say the user has "finished".  Perhaps it should list any missing data, only displaying Finished if all data is complete?

codingWithRach (approved):
But the actual functionality is great :-)

dblatcher:
> Some consideration needs to be given to what should be displayed on the Finish page for the Rendering Options wizard. Currently it says: Screenshot 2023-04-06 at 07 32 22
> 
> However, now that the user can skip around, there may be some pages that haven't been filled out, in which case it's not really accurate to say the user has "finished". Perhaps it should list any missing data, only displaying Finished if all data is complete?
> 
> (To keep this PR a manageable size, this could be considered in a separate PR)

👍  I've updated the content slightly to say that the options are set, but can be adjusted so seems less less "final". Does need a proper look at the messaging though.

Also, the link to the details page was broken since the routing change - I've fixed that!

PR #86, opened 2023-04-06
Dblatcher/newsletter categories

dblatcher:
## What does this change?

Adds a `NewsletterData.category` property (enum: ['article-based', 'fronts-based', 'manual-send', 'other']) and updates the fixtures, generator etc to include it.

Adds a step to the create/edit wizard to set the category.

Renders the category on the draft table, newsletters table and draft details.

Uses the category to determine whether to offer the user the "Rendering Options" wizard in the UI (IE only if the category is 'article-based')

## How to test

`npm run dev`

We will be able to use the category to better guide user in the creation/launch process - IE only asking the questions that are relevant to the category they are collecting. We can include fields the datamodel that are (for example) specific and optional for Fronts emails.

(long term goal) When the consumers of the newsletter data (frontend, identity etc) migrate from the legacy data model, the 'category' field will be available.

n/a

## Images

Screenshot 2023-04-06 at 12 34 43


libs/newsletter-workflow/src/lib/steps/newsletterData/categoryLayout.ts
```diff
@@ -0,0 +1,52 @@
...
+`.trim();
+
+const staticMarkdown = markdownTemplate.replace(
+	regExPatterns.name,
+	'the newsletter',
+);
+
+export const categoryLayout: WizardStepLayout = {
+	staticMarkdown,
+	label: 'production category',
```
codingWithRach, line 29: For consistency with other labels, should be 'Production Category'

libs/newsletter-workflow/src/lib/steps/newsletterData/createDraftNewsletterLayout.ts
```diff
@@ -20,7 +20,7 @@ The first step is to enter the name of your newsletter, for example **Down to Ea
 		next: {
 			buttonType: 'GREEN',
 			label: 'Next',
-			stepToMoveTo: 'dates',
+			stepToMoveTo: 'category',
```
codingWithRach, line 23: Also need to make the same change in editDraftNewsletterLayout.ts

codingWithRach (commented):
Looks really good - just need to change the Edit wizard to be consistent with the Create wizard

codingWithRach (approved):

PR #91, opened 2023-04-12
Dblatcher/generic wizards

dblatcher:
## What does this change?

Changes `WizardLayouts` and its sub-types (eg `AsyncValidator`, `WizardStepLayoutButton`) to  generic type with a parameter for the type of the "storageInstance" passed to the step event function (eg `executeStep`), so that Wizards are not tied to the `DraftStorage` class and can potentially interface with any set of external services.

Defines the existing layouts, "newsletterDataLayout" and "renderingOptionsLayout" as being `WizardLayout<DraftStorage>`

Refactors the code in the state-machine library to be generalised about the type of "storageInstance" the `WizardLayout` uses

## How to test
All functionality will be unaffected by this PR.

Will be possible to define wizards that take steps other than creating and editing draft newsletters - this in currently needed for the "launch newsletter wizard"

There is further refactoring needed to remove references to `DraftStorage` from the state-machine library's "utility" module.

### DIFF
--- a/libs/newsletter-workflow/src/lib/executeCreate.ts
+++ b/libs/newsletter-workflow/src/lib/executeCreate.ts
@@ -1,4 +1,7 @@
-import type { DraftNewsletterData } from '@newsletters-nx/newsletters-data-client';
+import type {
+	DraftNewsletterData,
+	DraftStorage,
+} from '@newsletters-nx/newsletters-data-client';
 import {
 	draftNewsletterDataToFormData,
 	formDataToDraftNewsletterData,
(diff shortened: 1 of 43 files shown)


codingWithRach:
I don't think this is quite there - I tried to invoke the Rendering Options wizard, and got the following errors:
Screenshot 2023-04-13 at 09 12 27
Screenshot 2023-04-13 at 09 12 20

codingWithRach:
Ditto on the setup wizard
Screenshot 2023-04-13 at 09 15 00

dblatcher:
> I don't think this is quite there - I tried to invoke the Rendering Options wizard, and got the following errors: Screenshot 2023-04-13 at 09 12 27 Screenshot 2023-04-13 at 09 12 20

Thanks - I've pushed a fix. The problem was that the utility functions in state-machine (which need refactoring out on the next PR) were using `storageInstance instanceof DraftStorage`, but `DraftStorage` is the abstract class.

Not sure how I had this working yesterday 🤔

phillipbarron (approved):
Tested locally - lgtm 👍

PR #93, opened 2023-04-17
Dblatcher/wizards specific initialstate and skip

dblatcher:
## What does this change?

Extends the `WizardStepLayout` to include optional `getInitialFormData` and `executeSkip` methods, allowing individual wizards to define how they handle these parts of the wizard lifecycle.

Refactored the existing code for this out of the state-machine library.

updates the existing newsletter-workflow steps to have the new methods as needed. 

Changed the 'rules' for skipping so that the current step must have an `executeSkip` method and the target must have the `canSkipTo` property set to true (enforced both in the state-machine and by whether the UI renders the skip button.

## How to test

Wizard should function as before.

State-machine no longer has any imports relating to draftNewsletters and can be considered generic, finishing the work started in https://github.com/guardian/newsletters-nx/pull/91

Future wizard layouts will be able to get the initial state form sources other than the draft storage

### DIFF
--- a/libs/newsletter-workflow/src/lib/getDraftFromStorage.ts
+++ b/libs/newsletter-workflow/src/lib/getDraftFromStorage.ts
@@ -1,0 +1,38 @@
+import { draftNewsletterDataToFormData } from '@newsletters-nx/newsletters-data-client';
+import type {
+	DraftStorage,
+	FormDataRecord,
+} from '@newsletters-nx/newsletters-data-client';
+import type { CurrentStepRouteRequest } from '@newsletters-nx/state-machine';
+import {
+	StateMachineError,
+	StateMachineErrorCode,
+} from '@newsletters-nx/state-machine';
+
+export const getDraftFromStorage = async (
+	requestBody: CurrentStepRouteRequest,
+	storageInstance: DraftStorage,
+): Promise<FormDataRecord> => {
+	const listId =
+		typeof requestBody.formData?.['listId'] === 'number'
+			? requestBody.formData['listId']
+			: undefined;
+	const existingItemId = listId ?? requestBody.id;
+
+	if (!existingItemId) {
+		// if there is no existingItemId, it is a new item so the state shoudl be empty
+		return {};
+	}
+	const idAsNumber = +existingItemId;
+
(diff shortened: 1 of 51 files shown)


libs/newsletter-workflow/src/lib/getDraftFromStorage.ts
```diff
@@ -0,0 +1,38 @@
...
+	storageInstance: DraftStorage,
+): Promise<FormDataRecord> => {
+	const listId =
+		typeof requestBody.formData?.['listId'] === 'number'
+			? requestBody.formData['listId']
+			: undefined;
+	const existingItemId = listId ?? requestBody.id;
+
+	if (!existingItemId) {
+		// if there is no existingItemId, it is a new item so the state shoudl be empty
```
phillipbarron, line 23:
```suggestion
		// if there is no existingItemId, it is a new item so the state should be empty
```

libs/state-machine/src/lib/handleWizardRequest.ts
```diff
@@ -50,7 +50,7 @@ export async function handleWizardRequestAndReturnWizardResponse<
 						wizardLayout,
 						genericInterface,
 				  )
-				: await setupInitialState(requestBody, genericInterface);
+				: await setupInitialState(requestBody, wizardLayout, genericInterface);
```
phillipbarron, line 53: not really part of this PR but perhaps there is a more descriptive name for `genericInterface`. `genericStorageInterface` maybe - something that clarifies its function.

phillipbarron (approved):
Looks good - though it would be good to talk through how storage is working. On an initial reading, that there is function to `getDraftFromStorage` seems a little unexpected. Ie, isn't that a  function of a  client or a property of a document? 
This is a  gap in my understanding and not a comment about the structure.
Tested locally and all seems to be behaving as expected

libs/state-machine/src/lib/handleWizardRequest.ts
dblatcher: sounds good - the naming does need to be addressed. We refer to this as a `storageInstance` in places, but potentially this interface won't necessary handle storage (or might handle storage but also something else) - I'll add an issue to look at it.

PR #94, opened 2023-04-18
Dblatcher/newsletter storage class

dblatcher:
## What does this change?

- Defines an abstract class for `NewsletterStorage` and `InMemoryStorage` implementation of it, to act in a similar way to the existing `DraftStorage`.
 - Creates an instance of the `InMemoryStorage` for the API to use instead of importing the static JSON. Updates the existing "api/newsletter" routes to use it. 
 - Adds an "api/newsletter/edit/:id" route to allow modification of existing newsletters.
 - Adds a page with a simple form for updating a subset of the newsletter's properties, linked from the newsletters table ( **NOTE: the form page was mostly added to make sure the update/modify sequence works. it could be replaced with a new wizard layout or alternative UI**)

## How to test

The "newsletters" pages will work as before, except the "edit" form can be used to change the data at run time. As the implementation is 'in memory' the changes will reset when the application restarts

This PR provides an interface for handling a mutable set of live newsletters. The same abstract class can be implemented to handle storage in S3 on CODE and PROD.

The NewsletterStorage class could be used within a Wizard as the draftStorage currently is - this will be necessary for the "launch" action

The `NewsletterStorage` has a `readByName` method and a `read` (by listI number)  - it when doing the s3 implementation, we should consider how to structure the bucket to search by either field without downloading all the items and reading the contents.

## Images

Screenshot 2023-04-18 at 14 05 37


apps/newsletters-ui/src/app/components/EditNewsletterForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+		}
+	};
+
+	return (
+		<Container maxWidth="lg">
+			<SimpleForm
+				title={`edit ${originalItem.identityName}`}
+				initalData={item}
+				submit={(modification) => {
+					void requestUpdate(modification);
```
codingWithRach, line 55: I know this form was only created as a proof of concept rather than necessarily being something that stays, but just in case it does stick around for a while it might be better if the Submit Form button returned to the list after saving the data.  Especially since if you Submit the form and then press Reset then nothing happens (which is perfectly logical, but might disappoint a user)

libs/newsletters-data-client/src/lib/newsletter-storage/NewsletterStorage.ts
```diff
@@ -14,14 +14,14 @@ export abstract class NewsletterStorage {
 		SuccessfulStorageResponse<NewsletterData> | UnsuccessfulStorageResponse
 	>;
 
-	abstract get(
+	abstract read(
```
codingWithRach, line 24: Good idea to go for CRUD naming convention, but it does beg the question of whether we should do the same with drafts e.g. change getDraftNewsletter to readDraftNewsletter etc?  Or even change getDraftNewsletter to read??

codingWithRach (approved):

libs/newsletters-data-client/src/lib/newsletter-storage/NewsletterStorage.ts
dblatcher: 👍 I was planning to do that at some point, but didn't want to have too big a diff on one PR - I'll add a health ticket.

apps/newsletters-ui/src/app/components/EditNewsletterForm.tsx
dblatcher: great idea - I've pushed a change for this. The `useNavigate` hook from react-router makes it easy to change pages.

dblatcher: oops - I didnt't realise the push didn't go through. I'll put it on another pr

PR #98, opened 2023-04-20
Allow skipping to section in data collection wizard

phillipbarron:
## What does this change?

- Allows skipping between steps. Requires that a  name be added as subsequent steps rely on a name being set.
- Removes the braze and identiy steos from the data collection wizard. These steps will be re-added in the lanch step in a later PR

## How to test

Chack out the brach & ensure it is possible to add update the steps out of sequence.

Always

## Images

hotStepper


codingWithRach (commented):
Looking good, but a few changes needed.

Some additional comments:
1. On the Create Wizard, it's possible to skip to the Enter Name step, but on the Edit Wizard, it's not possible to skip to the Change Name step
2. On the Edit wizard, the skips aren't enabled until after you've pressed Next to move off the Change Name step (inconsistent with the Create wizard)
3. On the Create wizard, if I leave the Name blank it correctly moans at me that the Name must not be empty.  If I then enter a name, it won't let me skip to another step - I have to press the Next button to move off the step first (this might be a restriction of the way it's all been constructed).
4. As per the chat messages: given that the IdentityName and Braze steps have been moved out of this wizard, the Ophan step probably should too.  This then opened up a can of worms about other pages that may or may not belong in this wizard.  However, this could  all probably be addressed in a separate PR :-)

libs/newsletter-workflow/src/lib/steps/newsletterData/ophanLayout.ts
```diff
@@ -63,4 +64,6 @@ export const ophanLayout: WizardStepLayout<DraftStorage> = {
 			stepToMoveTo: 'completeDataCollection',
```
codingWithRach, line 64: stepToMoveTo on line 53 needs to be updated now that braze has been moved out

libs/newsletter-workflow/src/lib/ececuteSkip.ts
```diff
@@ -0,0 +1,19 @@
+import type {DraftStorage} from "@newsletters-nx/newsletters-data-client";
+import type {AsyncExecution, WizardFormData, WizardStepData, WizardStepLayout} from "@newsletters-nx/state-machine";
+
+export const executeSkip: AsyncExecution<DraftStorage> = async (
```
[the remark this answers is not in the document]
phillipbarron: Opps - yup 👍

libs/newsletter-workflow/src/lib/steps/newsletterData/createDraftNewsletterLayout.ts
```diff
@@ -31,4 +32,6 @@ The first step is to enter the name of your newsletter, for example **Down to Ea
 	},
 	schema: formSchemas.startDraftNewsletter,
 	role: 'CREATE_START',
+	canSkipTo: true,
+	executeSkip: executeModify,
```
dblatcher, line 36: Skipping isn't working here - the "invalid or missing listId" message comes up.  This is because the "executeCreate" function (the 'next' button's `executeStep` ) needs to be be called to add the listId to the draftData before the data can be saved.

I think we could fix this by using "executeCreate" for executeSkip too - it would still require the user to enter a name before skipping. We need it to do this because "executeCreate" also generates the derived fields from the name currently. 

I think on a future PR, we should move the derived fields logic to the launch wizard - "executeCreate" would still set the "listId" (if not already set), but we wouldn't need to have the name set, so users could freely skip from this step.

Also, if we defer the derived fields until launch, they will be based on the name as it is at that point - currently, if you changed the name after setting it, the derived fields would still be based on the original name.

libs/newsletter-workflow/src/lib/steps/newsletterData/completeDataCollectionLayout.ts
```diff
@@ -36,7 +38,7 @@ export const completeDataCollectionLayout: WizardStepLayout<DraftStorage> = {
 		back: {
 			buttonType: 'RED',
 			label: 'Back',
-			stepToMoveTo: 'ophan',
+			stepToMoveTo: 'signUpEmbed',
 			executeStep: executeModify,
 		},
 		next: {
```
dblatcher, line 44: I don't think we need this step anymore, it's not in the index.ts for this folder. I think we can just delete the file.
If we do need it, the next step is still set to 'emailCentralProduction'

dblatcher (approved):
looking good - the createDraftNewsletterLayout isn't working exactly right, but I think that's one for another PR.

libs/newsletter-workflow/src/lib/steps/newsletterData/createDraftNewsletterLayout.ts
phillipbarron: Well spotted - thanks. Sounds sensible.

libs/newsletter-workflow/src/lib/steps/newsletterData/completeDataCollectionLayout.ts
phillipbarron: Updated - cheers

PR #100, opened 2023-04-20
Dblatcher/calculate draft progress

dblatcher:
## What does this change?

 - Calculates an approximate % score for the completeness of a draft to use in the draft table (replaces the random number). This is based on the number of validation errors that would prevent launch (0 errors = 100%), to filling out optional data won't change the score.
 - renders a launch button on the draft table (even if not ready to launch) that links to the wizard
 - renders a summary of the missing data on the draft details page.

## How to test

Visible in the UI:
 -  completion % will increase on the existing drafts as you provide values in the wizard
 - the issues list on the details page 

better visibility of the required data.

## Images
Screenshot 2023-04-20 at 17 05 17

Screenshot 2023-04-20 at 17 04 58


phillipbarron (approved):
LGTM 👍

PR #101, opened 2023-04-21
Dblatcher/state machine next function

dblatcher:
## What does this change?

changes the definition of a `WizardStepLayoutButton.stepToMoveTo` to be a `string | StepFindFunction`, where a `StepFindFunction` takes the wizardLayout and the current step are arguments to determine the stepId of the step to mve to.

Defines `StepFindFunction`s for going to the next or previous step in the sequence and uses them instead of the manually defined stepId's in the NewsletterData and RenderingOptions layouts.

This should reduce the amount of manual work needs to add or remove steps to a layout, since the 'next' and 'previous' ids no longer need to get maintained. 

Ids can still be defined manually as string where required - such as moving to a 'cancel' step or the "child" steps such as `editOphanLayout` or `editIdentityNameLayout`.

 
## How to test

Wizards still work. changing the order of the step in the definition of the index files for the layouts will re-order the steps in the UI without needing manual changes.

Defining and editting wizard layouts is less time comsuming and error prone.


libs/state-machine/src/lib/step-find-functions.ts
```diff
@@ -0,0 +1,86 @@
...
+		throw new StateMachineError(
+			'step is not part of wizard',
+			StateMachineErrorCode.NoSuchStep,
+			true,
+		);
+	}
+	return { allIds, allSteps, indexOfCurrentStep };
+};
+
+export const goToNextNormalStep: StepFindFunction = (wizard, step) => {
```
phillipbarron, line 29: these functions are returning the id of the next logical step - wondering if the name might reflect that? something like `getNextStepId` or similar

phillipbarron (approved):
Tested locally and working and it's a  big improvement. I am wondering how diffiult it might be to have a simpler naming style for the functions which determine the next or previous steps. This is not a  blocker to merging - this is an improvement on the existing implementation. We might want to, if possible, move to a  simpler `getNextStep`, `getPreviousStep` and paramterise to return the appropriate id.
We might even have a single function which takes a direction parameter ie:

```javascript
export const getNextStepId ({
  wizardLayout,
  wizardStep,
  isEditPath,
 direction
}) => {
....some codes.
}
```

something for later perhaps

This is great 👍

libs/state-machine/src/lib/step-find-functions.ts
dblatcher: 👍 that sounds good - it's a better description of function itself.

dblatcher:
We might want to, if possible, move to a simpler `getNextStep`, `getPreviousStep` and paramterise to return the appropriate id. We might even have a single function which takes a direction parameter ie:
> 
> ```js
> export const getNextStepId ({
>   wizardLayout,
>   wizardStep,
>   isEditPath,
>  direction
> }) => {
> ....some codes.
> }
> ```
> 

Parameterising sounds like a great idea. I think we'd have to use higher level functions since within state-machine the function will always be passed the same set of arguments - so something like

```typescript
export const getStepIdFor = (direction:'NEXT'|'BACK'|'START'):StepFindFunction => (wizard, step) => {

	//... some code 
	switch (direction) {
		//... some code using direction
	}

	return stepId
}
```

then in the step we would have :
```
{
  stepToMoveTo: getStepIdFor('NEXT'),
}
```

PR #107, opened 2023-04-27
Scroll fix & copy updates

phillipbarron:
Fixes scroll and updates copy

### DIFF
--- a/apps/newsletters-ui/src/app/Layout.tsx
+++ b/apps/newsletters-ui/src/app/Layout.tsx
@@ -12,7 +12,6 @@ const Frame = styled.div`
 	flex-direction: column;
 	height: 100vh;
 	box-sizing: border-box;
-	overflow: hidden;
 	align-items: stretch;
 
 	>
--- a/libs/newsletter-workflow/src/lib/steps/launchNewsletter/isReady.ts
+++ b/libs/newsletter-workflow/src/lib/steps/launchNewsletter/isReady.ts
@@ -45,7 +45,7 @@ export const isReadyLayout: WizardStepLayout<LaunchService> = {
 
 		return populated;
 	},
-	label: 'check if ready',
+	label: 'Check if ready',
 	buttons: {
 		cancel: {
 			buttonType: 'RED',
--- a/libs/newsletter-workflow/src/lib/steps/newsletterData/frequencyLayout.ts
+++ b/libs/newsletter-workflow/src/lib/steps/newsletterData/frequencyLayout.ts
@@ -11,15 +11,16 @@ const markdownTemplate = `
 # Specify the send frequency of {{name}}
 
 Specify how regularly the newsletter will be sent e.g.
-- every day
-- every weekday
-- weekly
-- fortnightly
-- monthly
+- Every day
+- Every weekday
+- Weekly
+- Fortnightly
+- Monthly
+- Monthly
 
 You can also add a different frequency e.g.
-- weekly during election season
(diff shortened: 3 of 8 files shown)


apps/newsletters-ui/src/app/Layout.tsx
```diff
@@ -12,7 +12,6 @@ const Frame = styled.div`
 	flex-direction: column;
 	height: 100vh;
 	box-sizing: border-box;
-	overflow: hidden;
```
phillipbarron, line 15: this is the scroll bit!

idrise (approved):

libs/newsletter-workflow/src/lib/steps/launchNewsletter/isReady.ts
```diff
@@ -45,7 +45,7 @@ export const isReadyLayout: WizardStepLayout<LaunchService> = {
 
 		return populated;
 	},
-	label: 'check if ready',
+	label: 'Check if ready',
```
codingWithRach, line 48: The labels have all been created in lowercase to make them consistent with the ones that aren't specifically specified because the autogenerated version is fine as it is (usually the case when it's a single word)

libs/newsletter-workflow/src/lib/steps/newsletterData/frequencyLayout.ts
```diff
@@ -11,15 +11,16 @@ const markdownTemplate = `
 # Specify the send frequency of {{name}}
 
 Specify how regularly the newsletter will be sent e.g.
-- every day
-- every weekday
-- weekly
-- fortnightly
-- monthly
+- Every day
```
codingWithRach, line 14: These are deliberately lowercase because of where they appear (see screenshots on this page), and the examples should reflect what we want the user to enter

libs/newsletter-workflow/src/lib/steps/newsletterData/thrasherLayout.ts
```diff
@@ -19,8 +19,6 @@ Should the single thrasher appear on app and web?
 
 Do you need one or more multi-thrashers i.e. sets where the thrasher for **{{name}}** is combined with thrashers for 2 or more other newsletters?
 
-***NEED TO DETERMINE THE BEST WAY TO CAPTURE THIS INFORMATION - CLEARLY A CHECKBOX IS INSUFFICIENT!***
```
codingWithRach, line 22: This was just a holding comment to remind us that the page was incomplete - but there is now a separate trello ticket for capturing this information

libs/newsletter-workflow/src/lib/steps/launchNewsletter/isReady.ts
codingWithRach: In any case, this step has been renamed in my WIP PR to add steps into the Launch wizard

codingWithRach: But then I hadn't spotted that somewhere along the line all the breadcrumbs have been changed so that we're no longer using defaults anywhere

libs/newsletter-workflow/src/lib/steps/newsletterData/frequencyLayout.ts
codingWithRach: Ignore this comment too - it's mixed case in the spreadsheet, so we're obviously lower-casing it before use (which is exactly as it should be)!

PR #109, opened 2023-04-28
Implement delete

phillipbarron:
## What does this change?

Adds a delete implementation for s3 newsletter storage
Fixes a bug where Modification errors in updates to launched newsletters were not being returned

## How to test

The Delete function for s3 is not exposed in the API as we have yet to decide what we are doing when we want to remove a launched newsletter. Added the delete implementation for completeness. It may change. Review the code and that's enough.

### DIFF
--- a/libs/newsletters-data-client/src/lib/newsletter-storage/InMemoryNewsletterStorage.ts
+++ b/libs/newsletters-data-client/src/lib/newsletter-storage/InMemoryNewsletterStorage.ts
@@ -118,11 +118,10 @@ export class InMemoryNewsletterStorage implements NewsletterStorage {
 		}
 
 		this.memory.splice(this.memory.indexOf(match), 1);
-		const response: SuccessfulStorageResponse<NewsletterData> = {
+		return Promise.resolve({
 			ok: true,
-			data: match,
-		};
-		return Promise.resolve(response);
+			data: `newsletter #${listId} deleted`,
+		} as SuccessfulStorageResponse<string>);
 	}
 
 	list() {
--- a/libs/newsletters-data-client/src/lib/newsletter-storage/NewsletterStorage.ts
+++ b/libs/newsletters-data-client/src/lib/newsletter-storage/NewsletterStorage.ts
@@ -43,7 +43,7 @@ export abstract class NewsletterStorage {
 	abstract delete(
 		listId: number,
 	): Promise<
-		SuccessfulStorageResponse<NewsletterData> | UnsuccessfulStorageResponse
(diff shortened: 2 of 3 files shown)


libs/newsletters-data-client/src/lib/newsletter-storage/NewsletterStorage.ts
```diff
@@ -42,9 +42,7 @@ export abstract class NewsletterStorage {
 
 	abstract delete(
 		listId: number,
-	): Promise<
-		SuccessfulStorageResponse<NewsletterData> | UnsuccessfulStorageResponse
-	>;
+	): Promise<SuccessfulStorageResponse<string> | UnsuccessfulStorageResponse>;
```
dblatcher, line 45: I get that fetching the data from s3 before deleting it is a pain, so returning a copy of the deleted `NewsletterData` is probably overkill - but just returning a string might make it harder for the consumer of the storage to interpret and react to  the responses - I'd tend to return a simple object like:
```
type DeletionConfirmation = {
  message: string;
  listId: number;
}
```

libs/newsletters-data-client/src/lib/newsletter-storage/InMemoryNewsletterStorage.ts
```diff
@@ -118,11 +118,10 @@ export class InMemoryNewsletterStorage implements NewsletterStorage {
...
 		this.memory.splice(this.memory.indexOf(match), 1);
-		const response: SuccessfulStorageResponse<NewsletterData> = {
+		return Promise.resolve({
 			ok: true,
-			data: match,
-		};
-		return Promise.resolve(response);
+			data: `newsletter #${listId} deleted`,
+		} as SuccessfulStorageResponse<string>);
 	}
```
dblatcher, line 125: as below - I'd suggest returning an object rather than a string just to make the responses easier to handle.

libs/newsletters-data-client/src/lib/newsletter-storage/s3-newsletter-storage.ts
```diff
@@ -97,18 +98,30 @@ export class S3NewsletterStorage implements NewsletterStorage {
...
+		} catch (error) {
+			return {
+				ok: false,
+				message: `failed to delete newsletter with id ${listId}`,
+				reason: StorageRequestFailureReason.S3Failure,
+			};
+		}
+		return {
+			ok: true,
+			data: `newsletter with id ${listId} deleted`,
```
dblatcher, line 123: same point about the data type again.

dblatcher (approved):
Looks good - one suggestion about the return value for the delete method.

PR #111, opened 2023-05-01
Replaced Wizard buttons with styled MUI buttons...

phillipbarron:
## What does this change?

Updates the Wizard buttons and replaces to `Red`, `Green` types with some action themed type:

```javascript
export type WizardButtonType =
	| 'NEXT'
	| 'PREVIOUS'
	| 'EDIT'
	| 'SKIP'
	| 'LAUNCH'
	| 'CANCEL';
```

This PR also removes refs to `Demo` from buttons and routes with the exception of `Demo form` which really does look like a demo. 

## How to test

Check the code out and verify that the existing functionality is as before. The behaviour should not have changed - only some of the routes.

## Images

Screenshot 2023-05-01 at 21 26 56

Screenshot 2023-05-01 at 21 32 14


dblatcher:
looks great!

PR #112, opened 2023-05-03
Add derived values steps to launch wizard

codingWithRach:
## What does this change?

The steps for viewing and (optionally) editing the derived values (identity name, braze variables, ophan variables) were removed from the initial newsletterData wizard: this PR adds those pages into the Launch wizard.

## How to test

Run `npm run dev`
From the list of drafts, select a draft and enter all mandatory fields (5770 Avon Recumbent is already fully populated)
Click the Launch button and navigate through the wizard
Ensure that it works both if selecting Next all the way through, and also if selecting any of the Edit options
Check that editing data on other wizards still works (as a shared function has been changed)

Data changes, if made, are saved
After pressing Launch at the end of the wizard, the newsletter is moved from the list of Drafts to the list of Launched

This wizard does not allow the user to skip steps
Furthermore, a Back button is available only on the pages showing derived fields (otherwise, it is not possible to navigate backwards and forwards)
The user needs to complete the Launch wizard steps in order

The derivation of the fields from the newsletter name needs to be moved from its current position, but this will be addressed in a separate PR

## Images

(Note that this was recorded before merging from main)

https://user-images.githubusercontent.com/74301289/235904366-896601d2-6ed0-44d2-a170-85112d85a293.mov

## Accessibility


libs/newsletter-workflow/src/lib/steps/launchNewsletter/doLaunchLayout.ts
```diff
@@ -0,0 +1,44 @@
+import type { LaunchService } from '@newsletters-nx/newsletters-data-client';
+import type { WizardStepLayout } from '@newsletters-nx/state-machine';
+import { executeLaunch } from '../../executeLaunch';
+import { getStringValuesFromRecord } from '../../getValuesFromRecord';
+import { regExPatterns } from '../../regExPatterns';
+
+// TODO - this page is a placeholder until we implement automation
+// I have added it so that both Ophan and Edit Ophan point to a single place from which to call the Launch functionality
+const markdownTemplate = `
```
dblatcher, line 9: is this comment still current? I can remember if I copied it from when this was part of the main Draft Wizard.

libs/newsletter-workflow/src/lib/steps/launchNewsletter/doLaunchLayout.ts
```diff
@@ -0,0 +1,44 @@
...
+	},
+	label: 'Launch',
+	buttons: {
+		cancel: {
+			buttonType: 'CANCEL',
+			label: 'Cancel',
+			stepToMoveTo: 'cancel',
+		},
+		next: {
+			buttonType: 'NEXT',
```
dblatcher, line 38: It won't have any visual effect yet, but we should probably use the "LAUNCH" type on this button.

dblatcher (approved):

libs/newsletter-workflow/src/lib/steps/launchNewsletter/doLaunchLayout.ts
codingWithRach: Yes, I added that comment as part of this PR.  I don't expect this page to stay as it is, but it seemed better than having a launch button on both the Ophan and Edit Ophan steps, given that additional pages will inevitably be inserted between those and the end of the wizard (either in the mvp or afterwards)

libs/newsletter-workflow/src/lib/steps/launchNewsletter/doLaunchLayout.ts
codingWithRach: Good spot!

PR #113, opened 2023-05-03
Default production category to blank

codingWithRach:
## What does this change?

Defaults the Production Category to blank on the create newsletter wizard, so that users are forced to specifically select the category rather than just accepting the default

## How to test

`npm run dev`
Select the Create Newsletter Wizard
Enter a name and click Next
Ensure that
- no production category is displayed by default
- it's not possible to click next without selecting a production category

## Images

Screenshot 2023-05-03 at 15 14 10

Screenshot 2023-05-03 at 15 25 08

## Accessibility

### DIFF
--- a/libs/newsletters-data-client/src/lib/newsletter-data-type.ts
+++ b/libs/newsletters-data-client/src/lib/newsletter-data-type.ts
@@ -72,7 +72,7 @@ export const thrasherOptionsSchema = z.object({
 export type ThrasherOptions = z.infer<typeof thrasherOptionsSchema>;
 
 export const newsletterCategoriesSchema = z
-	.enum(['article-based', 'fronts-based', 'manual-send', 'other'])
+	.enum(['', 'article-based', 'fronts-based', 'manual-send', 'other'])
 	.describe('production category');
 export type NewsletterCategory = z.infer<typeof newsletterCategoriesSchema>;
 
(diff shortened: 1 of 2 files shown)


libs/newsletters-data-client/src/lib/newsletter-data-type.ts
```diff
@@ -72,7 +72,7 @@ export const thrasherOptionsSchema = z.object({
 export type ThrasherOptions = z.infer<typeof thrasherOptionsSchema>;
 
 export const newsletterCategoriesSchema = z
-	.enum(['article-based', 'fronts-based', 'manual-send', 'other'])
+	.enum(['', 'article-based', 'fronts-based', 'manual-send', 'other'])
```
dblatcher, line 75: We do this for other fields already so not a thing for this PR, but it's not ideal to be adding the empty string to the enum since its part of the schema for the actual `NewsletterData` type - I.E. this makes an empty string a valid value for `NewsletterData.category` (but the validation on the step will prevent it actually being selected). 

I'll put in a health ticket to look at this point. 🤔

dblatcher (approved):

libs/newsletters-data-client/src/lib/newsletter-data-type.ts
codingWithRach: That's a very good point - I was making it consistent with existing steps (and I think I'd done it that way originally as a workaround for it being a mandatory field), but if there's a better way to tackle it that's definitely worth looking at separately

PR #114, opened 2023-05-04
Dblatcher replace legacy details component

dblatcher:
## What does this change?

Replaces the `LegacyNewsletterDetail` component (that used a `LegacyNewsletter` as the prop)  with a new `NewsletterDataDetails` component.

## How to test

`npm run dev`
Open a launched newsletter by clicking the link on the identity name in the table.

The details page is not limited to displaying legacy values and makes use of MUI components to make the large amount of data easier to digest / navigate. Users can expand the carousels to find the category of data they need.

The page deliberately does not present every data point in the `NewsletterData` to the user on the main view, only those expected to be relevant. The 'View Raw Data' dialog presents the full set if needs and includes a 'developer friendly' feature for copying the whole JSON data to your clipboard.

The change is only visual / display, no risks anticipated

## Images

|before|after|
|----|----|
|Screenshot 2023-05-04 at 12 12 14| Screenshot 2023-05-04 at 12 09 59|


phillipbarron (approved):
This looks great. 

One minor suggestion is that we use the same colours and border-radius for the buttons that we are using in the Wizard buttons to keep it consistent - see https://github.com/guardian/newsletters-nx/blob/main/apps/newsletters-ui/src/app/components/Wizard.tsx#L133

dblatcher:
> 

👍 will do.

PR #116, opened 2023-05-04
Dblatcher no edit dialogue if one option

dblatcher:
## What does this change?

The draft table will render a link button going directly to the Edit Draft wizard when that is the only relevant wizard, instead of opening the dialog with the options.

In the current state of the application, that means direct link button will be rendered if the draft is not 'article-based' since that is the only condition where there is an extra wizard. If we add more wizards and update the `getEditDraftWizardLinks` function, the the logic in the draft table will still work (IE render the link dialog if there are more than one option, otherwise go to the only option directly).

https://trello.com/c/VROy98ZP/399-if-email-is-not-article-based-miss-out-modal-dialog-on-edit-option

## How to test

go to the drafts table - the edit buttons for the none "article rendering" drafts will go straight to the newsletterData wizard.


phillipbarron (approved):
lgtm

PR #117, opened 2023-05-04
Dblatcher collect grouping

dblatcher:
## What does this change?

Adds a drop down for the 'group' property on the same step as the 'pillar' option.
removes the default value for this field from the 'draft to newsletter' function - value should no longer be defaulted when trying to launch since it can (and should) be entered.

The 'group' property is a string in the `newsletterDataSchema` - this makes sense there isn't an external constraint definition what a valid 'group' is (the value determined which subheading the newsletter appears under in the all newsletters page), but we don't want users to have to type the value in, or create lots of new groups which could mess up the layout of the page, so the formSchema uses an enum instead of picking the field from the `newsletterDataSchema`.

This would mean that we need a PR to change the list of groups in available in the wizard, but we are unlikely to be adding new groups frequently.
https://trello.com/c/r5WGuCaK/342-collect-grouping-for-all-newsletters-page

## How to test

create a draft or edit an existing one.

## Images

|before|after|
|---|---|
| Screenshot 2023-05-04 at 15 51 58|Screenshot 2023-05-04 at 15 49 54 |

### DIFF
--- a/apps/newsletters-api/src/app/responses.ts
+++ b/apps/newsletters-api/src/app/responses.ts
@@ -1,7 +1,7 @@
 import { StorageRequestFailureReason } from '@newsletters-nx/newsletters-data-client';
 import type { ApiResponse } from '@newsletters-nx/newsletters-data-client';
 
-// TO DO - make the error response type generic
+// TODO - make the error response type generic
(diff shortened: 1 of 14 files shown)


phillipbarron (approved):
LGTM

PR #118, opened 2023-05-04
Dblatcher rename draft methods

dblatcher:
## What does this change?
Renames the method in the draft storage class to match CRUD naming instead of being unnecessarily long.

(https://trello.com/c/x4Os75Ok/383-rename-methods-in-draftstorage-class-to-match-the-crud-naming)

## How to test

Will run as before - no change in behaviour


libs/newsletters-data-client/src/lib/draft-storage/DraftStorage.ts
```diff
@@ -11,31 +11,31 @@ export type DraftWithId = DraftNewsletterData & { listId: number };
...
 
-	abstract modifyDraftNewsletter(
+	abstract update(
 		draft: DraftWithId,
 	): Promise<
 		SuccessfulStorageResponse<DraftWithId> | UnsuccessfulStorageResponse
 	>;
 
-	abstract deleteDraftNewsletter(
+	abstract deleteItem(
```
codingWithRach, line 32: Is there a reason why this is `deleteItem` rather than `delete` e.g. does `delete` get in a mess with reserved words or something?

dblatcher: exactly - "delete" is a [javascript operator] (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete) so it seemed better to not use it as a method name. It probably would have been fine in most cases, but if you ever wanted to do something like ```const {create, read, update, delete} = launchService``` it would break.

PR #119, opened 2023-05-05
Dblatcher indicate step completion

dblatcher:
## What does this change?

Removes the empty strings in the enum fields of the `newsletterDataSchema` - these were used so the user was not presented with the first item on the enum "pre-filled". Instead, changed the `SelectInput` components within `SchemaForm` to default the value to `undefined` when rendering a form element for an enum field.

The `StepNav` component now checks if each step is 'complete' based on the `formData` prop and, if the `WizardLayout` is configured to render the indicator, adds captions to the `Step` representing the `WizardStepLayout` to  tell the user if they are complete.

https://trello.com/c/KnjAJC3q/398-change-step-nav-to-give-visual-indication-of-whether-a-step-has-been-completed
https://trello.com/c/knzi0QIW/427-find-a-better-way-to-have-enum-dropdown-start-blank-rather-than-defaulting-to-the-first-item

## How to test

Go through the newsletterData and renderOptions wizards - the step labels will show if the steps are completed or not. The indicators will update as you change steps. Steps without a form (eg 'finish') will not have a caption.

Note that the Launch wizard is configured to have `indicateStepsComplete` false so will look the same.

Easier for users to see which steps need attention.
Schema no longer allows empty strings in the enum fields.

## Images

Screenshot 2023-05-05 at 15 06 42

### DIFF
--- a/apps/newsletters-ui/src/app/components/SchemaForm/SelectInput.tsx
+++ b/apps/newsletters-ui/src/app/components/SchemaForm/SelectInput.tsx
@@ -4,7 +4,7 @@ import type { FunctionComponent } from 'react';
 import { defaultFieldStyle } from './styling';
 import type { FieldProps } from './util';
 
-const undefToken = '_____';
+const EMPTY_STRING = '';
 
 export const SelectInput: FunctionComponent<
 	FieldProps & {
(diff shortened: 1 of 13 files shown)


apps/newsletters-ui/src/app/components/SchemaForm/SelectInput.tsx
```diff
@@ -15,25 +15,22 @@ export const SelectInput: FunctionComponent<
...
-					label={label}
-					onChange={handleChange}
-				>
-					{optional && <MenuItem value={undefToken}>Select value</MenuItem>}
+				<InputLabel id={`select-input-label-${label}-${options.toString()}`}>
+					{label}
+				</InputLabel>
+				<Select value={valueAsString} label={label} onChange={handleChange}>
+					{optional && <MenuItem value={EMPTY_STRING}>[none]</MenuItem>}
 					{options.map((option) => (
```
idrise, line 34: way nicer!

apps/newsletters-ui/src/app/components/StepNav.tsx
```diff
@@ -31,6 +84,31 @@ export const StepNav = ({
...
+		}, {});
+
+		setCompletionRecord(list);
+	};
+
+	// On the initial render, the completionRecord is set to {}
+	if (!Object.keys(completionRecord).length) {
+		updateCompletion();
+	}
+
```
idrise, line 105: little tip I got taught at a previous company that had a typedoc only policy...
It's quite nice and I use it quite a bit.

const isCompletionRecordPopulated = !Object.keys(completionRecord).length;
if (isCompletionRecordPopulated) {
  updateCompletion();
}
This is not a request to change, just sharing something I learned from someone smarter than me.

libs/state-machine/src/lib/getStepList.ts
```diff
@@ -26,16 +29,25 @@ export const getStepperConfig = (wizard: WizardLayout): StepperConfig => {
...
 					canSkipTo: step.canSkipTo,
 					canSkipFrom: !!step.executeSkip,
+					schema: step.schema,
 				},
 			];
 		},
 		[],
 	);
 
+	// TO DO - these should really be properties of the WizardLayout, rather than
```
idrise, line 39: If you rename TO DO as TODO I think vscode picks it up as a work list with an extension

image

idrise (approved):
Nice one!

libs/state-machine/src/lib/getStepList.ts
dblatcher: Thanks!  we have a quite a lot of "TO DO"'s in the repo - I'll add a heath ticket to make them all "TODO". 

```
// TODO make this TO DO a TODO
// TO DO - fix this
```

PR #124, opened 2023-05-10
Dblatcher remove emailembed schema

dblatcher:
## What does this change?

Removes the `emailEmbed` structure from the `NewsletterDataSchema` and derives it from the other fields when converting to a `LegacyNewsletter`.

Updates fixtures and generators to match.

See corresponding change in migration tool:
https://github.com/guardian/newsletters-migration-tool/pull/11/files

## How to test

features work as before in the UI.
The legacy data is still correctly populated.

The data model is simpler and no longer required to hold redundant legacy fields, which make it hard to determine what data is relevant.


phillipbarron (approved):
Awesome

PR #126, opened 2023-05-10
Dblatcher neat stepnav

dblatcher:
## What does this change?

Uses props and some custom global CSS to override the styling on the `StepButton`s in our the `StepNav` to give each button a uniform width, left alignment and a hover state.

## How to test

Change is purely visual - try some wizards in the UI

Nav looks neater when wrapping onto multiple lines.

Using global css with classname is a bit of a hack - the pattern is suggested as an option in the docs (https://mui.com/material-ui/api/step-button/#css) but isn't ideal. Relying on it too much would not be maintainable.

## Images

|before|after|
|---|---|
| Screenshot 2023-05-10 at 15 28 36| Screenshot 2023-05-10 at 15 28 23|

### DIFF
--- a/apps/newsletters-ui/index.html
+++ b/apps/newsletters-ui/index.html
@@ -18,6 +18,18 @@
 			body {
 				margin: 0;
 			}
+			.left-aligned-step-button b {
+				text-decoration: underline;
+			}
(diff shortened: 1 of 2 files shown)


phillipbarron (approved):
LGTM

PR #127, opened 2023-05-10
Dblatcher change edit form

dblatcher:
## What does this change?

The edit (launched) newsletter form now 
 - allows the user to edit the copy fields etc that editorial are more likely to want to update 
 - uses a dialog to show when the page is waiting for a response and will not submit another request until the response to the first is recieved and handled.

## How to test

http://localhost:4200/newsletters/edit/blue-empowering

Gives a better demo of a feature of the tool


apps/newsletters-ui/src/app/components/EditNewsletterForm.tsx
```diff
@@ -12,15 +18,29 @@ interface Props {
 	originalItem: NewsletterData;
 }
 
+const sleep = async (ms: number) =>
```
phillipbarron, line 21: Tested locally without this and seemed to work - not clear what this is doing for us? The Patch having completed should be suffient to tell us the update has finished

phillipbarron (commented):
This is a great addition but I wonder about the arbitarty 500 ms wait - I'm sure there is something I am missing.

Not part of this PR but I notice that the button styles do not match the Wizard button styles. Would be nice to align them here too.

I am assuming that we are using the modal to prevent access to the form during update. I wonder whether it might be less jaring an experience to update the simple form to have a `updating` parameter or simlar that we can use to disable inputs / show loading state in buttons. 
I'm no UX expert but I understand in general that modals are to be used sparingly as it can feel like an interruption to workflow / cookie like / ad-like attention grabber.

If that feels like to great a diversion, perhaps we could write a ticket together splitting that behaviour out?

apps/newsletters-ui/src/app/components/EditNewsletterForm.tsx
dblatcher: I originally put the delay in to test that the loading/waiting state would work when the update wasn't instant. But since the behaviour I was asked to put in on the last PR for this form was to navigate to the list view when the user submits the change, it struck me that there needs to be enough of a delay for the user to see the message that says what is happening.
I'm not sure that navigating to the list view is the best UX, actually. If they stay on the same page, they can just see the messaging that says 'newsletter updated'.

dblatcher:
> Not part of this PR but I notice that the button styles do not match the Wizard button styles. Would be nice to align them here too.

I'd like to do that globally with a MUI ThemeProvider at some point - hard coding the styling is going to get us unstuck eventually.

> I am assuming that we are using the modal to prevent access to the form during update. I wonder whether it might be less jaring an experience to update the simple form to have a `updating` parameter or simlar that we can use to disable inputs / show loading state in buttons. I'm no UX expert but I understand in general that modals are to be used sparingly as it can feel like an interruption to workflow / cookie like / ad-like attention grabber.

That sounds good - an optional `isDisabled` or `isLoading` prop would be a nice addition to the `SimpleForm`. I'll put that in and use something subtler than the modal dialog or the messaging.

phillipbarron:
apologies - forgot to add a ✅ last time

PR #131, opened 2023-05-12
Adding forwarding rule to api listener to block traffic 

phillipbarron:
## What does this change?

We should prevent unauthenicated traffic from reaching the Readonly application to protect us from DDoS attacks

This change adds a rule to the listened for the read-only api whcg checks for a header and value. Where these are present, the request is forwarded to the application

We will automatically add these header for requests from Fastly. 

## How to test

Try accessing the applications - requests should fail in the absence of the header and value described in the stacks.

This PR is a WIP - We might choose to move the header name in to the Param store too (and change it from its current value)

Request without header fail, requests with succeed

always

## Accessibility

### DIFF
--- a/cdk/lib/newsletters-tool.ts
+++ b/cdk/lib/newsletters-tool.ts
@@ -11,7 +11,9 @@ import { GuHttpsEgressSecurityGroup } from '@guardian/cdk/lib/constructs/ec2';
 import { type App, Duration, SecretValue } from 'aws-cdk-lib';
 import { InstanceClass, InstanceSize, InstanceType } from 'aws-cdk-lib/aws-ec2';
 import {
+	ApplicationListenerRule,
 	ListenerAction,
+	ListenerCondition,
 	UnauthenticatedAction,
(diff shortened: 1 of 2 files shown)


cdk/lib/newsletters-tool.ts
```diff
@@ -137,6 +139,31 @@ export class NewslettersTool extends GuStack {
...
+		const readOnlyEndpointApiKey = StringParameter.valueForStringParameter(
+			this,
+			readOnlyEndpointApiKeyParam,
+		);
+
+		new ApplicationListenerRule(this, 'ReadOnlyApiHeaderRule', {
+			listener: ec2AppApi.listener,
+			priority: 1,
+			conditions: [
+				ListenerCondition.httpHeader('X-Gu-API-Key', [readOnlyEndpointApiKey]),
```
phillipbarron, line 152: What is an appropriate format for this and might this be a parameter too? ie, we would hide both key and value in the repo

dblatcher (approved):

PR #132, opened 2023-05-15
Dblatcher change  edit form update

dblatcher:
## What does this change?

 - adds `isDisabled` and `message` props to the `SimpleForm`
 - changes the EditNewsletters form:
   - no fake delay
   - no navigation back to the list after the change is complete
   - use the new props on the SimpleForm instead of a Dialog modal


phillipbarron (approved):
LGTM 👍

codingWithRach:
Not directly related to this PR, but wondering why `email confirmation' is included on the form?  Isn't this a field that we're not collecting in the wizard because it's not used anywhere (exists historically in spreadsheet, but blank for all newsletters in PROD)?  Or have I misremembered that?

codingWithRach:
I know I'm the odd one out here, but I do find the interface confusing (from the user's perspective).  If you make a change, and Update Newsletter and then Reset it now does reset it to what it was before you updated.  If you say Back to List and then edit again, the Reset hasn't stuck (because you didn't say Update Newsletter after doing Reset), but there's no warning about unsaved changes when you click Back to List.  But I think I have an old-fashioned way of looking at things (as none of my previous development experience has involved a browser, so it was dialog boxes all the way!)

dblatcher:
> Not directly related to this PR, but wondering why `email confirmation' is included on the form? Isn't this a field that we're not collecting in the wizard because it's not used anywhere (exists historically in spreadsheet, but blank for all newsletters in PROD)? Or have I misremembered that?

It is a feature that still works - setting the value to 'true' will mean that identity will send a confirmation email ("please click this link to confirm you want to sign up to *newsletter name*) instead of signing-up directly. However, you're probably right that it’s not a feature we want to expose in the UI.  👍  I'll take it off.

Even if a need to turn it on arose, its unlikely editorial would need to toggle it on demand.

dblatcher:
> I know I'm the odd one out here, but I do find the interface confusing (from the user's perspective). If you make a change, and Update Newsletter and then Reset it now does reset it to what it was before you updated. If you say Back to List and then edit again, the Reset hasn't stuck (because you didn't say Update Newsletter after doing Reset), but there's no warning about unsaved changes when you click Back to List. But I think I have an old-fashioned way of looking at things (as none of my previous development experience has involved a browser, so it was dialog boxes all the way!)

All good points - we could do with  some UX input on the product.

PR #133, opened 2023-05-15
Dblatcher theming

dblatcher:
## What does this change?

 - sets up a `ThemeProvider` in the root component applying the following to MUI components:
   - a "primary" color theme based on the source palette "brand" values (and the shade of blue we were manually setting) - not very noticeable as the MUI default primary theme happens to use similar shades of blue
   - a "secondary" color theme based on source  palette "news" values (we don't use secondary colors much)
   - borderRadius:0 for buttons by default
   - typography values, for h1 ("newsletters" in app header nav) , h2 (newsletter and draft summary pages)
 - moves the css for the StepNav buttons out of the html template and into the existing `DefaultStyles` component
 - replaces some of the custom styling for forms using Mui component props. 
 - removes the manual styling on buttons that is not handled by the theme,

## How to test

Adjust the values in apps/newsletters-ui/src/app-theme.ts to see how they impact the appearance of the app.

Note that not every component uses MUI - those that don't will not be effected by the theme (sidenote - changing the font-family in apps/newsletters-ui/src/app/components/DefaultStyles.tsx is a good way to see which components are not usin MUI)

Easier to consistently apply styling throughout the app. Should make design feedback easier to implement

Does tie use closer to MUI.

## Images
Screenshot 2023-05-15 at 17 21 06

### DIFF
--- a/apps/newsletters-ui/index.html
+++ b/apps/newsletters-ui/index.html
@@ -18,18 +18,6 @@
 			body {
 				margin: 0;
 			}
-			.left-aligned-step-button b {
-				text-decoration: underline;
-			}
-			.left-aligned-step-button > .MuiStepLabel-horizontal {
-				flex: 1;
-				text-align: left;
-				min-height: 4rem;
-				transition: background-color 0.5s;
-			}
-			.left-aligned-step-button > .MuiStepLabel-horizontal:hover {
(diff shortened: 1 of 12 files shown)


phillipbarron (approved):
This is great - big improvement

PR #136, opened 2023-05-16
add page with list of email rendering templates

dblatcher:
## What does this change?

Adds a page under "email-templates" with a list of links to the chromatic stories for the templates implemented on email-rendering. The data comes from http://email-rendering.guardianapis.com/info/templates.

## How to test

http://localhost:4200/templates

Provides a reference point for users to see the implemented designs.

The link would be broken if we create a new template in email rendering, but don't add it to the primary template story.

We could fix this by adding the storybook url to the Template data in email rendering and including it in the JSON (if the url is undefined, the page would not render a link), but that would be extra housekeeping. 

## Images

Screenshot 2023-05-18 at 17 18 06


apps/newsletters-ui/src/app/components/TemplateList.tsx
```diff
@@ -0,0 +1,75 @@
...
+			return `${EMAIL_RENDERING_CHROMATIC_MAIN}/?path=/story/playground-content--default&args=editBody:false;template:first-edition`;
+		default:
+			return `${EMAIL_RENDERING_CHROMATIC_MAIN}/?path=/story/templates-primarytemplate--${template.id}`;
+	}
+};
+
+export const TemplateList = ({ templates }: Props) => {
+	return (
+		<Container maxWidth="lg">
+			<Typography variant="h2">email-rendering Templates</Typography>
```
phillipbarron, line 33:
```suggestion
			<Typography variant="h2">Email rendering templates</Typography>
```

phillipbarron (approved):
This i sa great improvement. I have one suggestion which is that we use that same font everywhere in the app -  The typeface in the header on the new page is distinctly different from the rest of the text on the page and in the Nav

PR #137, opened 2023-05-16
Collect data for multi thrashers

codingWithRach:
## What does this change?

The Thrashers step of the wizard has been renamed Single Thrasher, and the checkbox indicating the requirement for multi-thrasher(s) has been removed

An additional Multi Thrashers step has been inserted, allowing the definition of one or more multi-thrashers

## How to test

Run `npm run dev`
Select Edit for a newsletter in the list of Drafts and navigate through the Newsletter Set-up wizard (this wizard will be displayed automatically if not an article-based newsletter)

All data is now being collected in a single place to enable a developer to implement the new email-rendering template and thrashers

There are several restrictions to the way this has been implemented, which may need to be addressed in future:
- the interface assumes all multi-thrashers will be triple thrashers.  For the vast majority this is the case, but occasionally more than 3 newsletters are required
- the individual newsletters within a triple-thrasher are currently specified via a free-entry text field.  Ideally this would be replaced by a drop-down comprising the names of all Launched newsletters plus the name of the Draft newsletter currently being created.

The above functionality is beyond the scope of the mvp

## Images

Before
thrasherLayout - before

Single Thrashers after
Screenshot 2023-05-16 at 16 40 19
 
Multi Thrashers after
Screenshot 2023-05-16 at 16 41 05

## Accessibility


dblatcher (approved):
Looks good!
For a future version of the tool, I think we should consider moving the thrasher questions to a their own wizard - there's also an argument for modeling them separately to the Newsletters themselves - IE having a list of thrashers that is separate to the list of newsletters, but can refence the Newsletters.

PR #138, opened 2023-05-16
derive name-based fields during launch

dblatcher:
## Accessibility

### DIFF
--- a/apps/newsletters-api/src/services/storage/inMemoryStorageInstance.ts
+++ b/apps/newsletters-api/src/services/storage/inMemoryStorageInstance.ts
@@ -3,18 +3,14 @@ import { InMemoryDraftStorage } from '@newsletters-nx/newsletters-data-client';
 export const makeInMemoryStorageInstance = () =>
 	new InMemoryDraftStorage([
 		{
-			identityName: 'avon-recumbent',
 			name: 'Avon Recumbent',
-			category: 'article-based',
+			category: 'fronts-based',
 			signUpDescription:
 				'Quod nobis aperiam aperiam ipsum nobis officiis explicabo molestias quibusdam.',
 			frequency: 'Weekly',
 			status: 'paused',
 			restricted: false,
 			emailConfirmation: false,
-			brazeNewsletterName: 'Editorial_AvonRecumbentUs',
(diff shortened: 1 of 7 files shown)


codingWithRach:
This is good.  Ideally we also need to validate at the point where the derivation is done.  It's a bit contrived, but during the launch process for Avon Recumbent I changed the identity name to background-hatchback (which is of course allowed as Background Hatchback is a draft, and so no identityName has yet been derived).  I then filled in the missing fields for Background Hatchback and launched that, accepting the default identity name.  The duplicate was only detected at the point where the final Launch button was pressed.  It would be better (if not too much of a faff) if we don't allow the duplicate to be set up in the first place (although this is admittedly an unlikely scenario in practice)

dblatcher:
> This is good - your approach is better than the one I'd been attempting. I couldn't work out how to save the derived values in the Launch wizard, but you aren't doing that - which makes a lot of sense given that the user might still cancel out of the Launch and change the newsletter name.
> 
> Ideally we also need to validate at the point where the derivation is done. It's a bit contrived, but during the launch process for Avon Recumbent I changed the identity name to background-hatchback (which is of course allowed as Background Hatchback is a draft, and so no identityName has yet been derived). I then filled in the missing fields for Background Hatchback and launched that, accepting the default identity name. The duplicate was only detected at the point where the final Launch button was pressed. It would be better (if not too much of a faff) if we don't allow the duplicate to be set up in the first place (although this is admittedly an unlikely scenario in practice)
> 
> In a similar scenario to the above, if I click the Edit button for the identity name on Background Hatchback, it correctly validates that background-hatchback already exists. But I can click Back and then Next (i.e. skipping Edit) and it will allow it until the final Launch button.

Cunning! i could imagine this coming up if someone created a new newsletter that had the same name as an existing or cancelled one, but was still meant to be a 'new' newsletter (like when we did the trial for the auto-subscribed newsletter recently, "saturday-roundup-trial").
I'll have a go at addressing this.

dblatcher:
@codingWithRach I've pushed some changes to make sure the "recommended" derived values aren't already used by exsiting launched newsletters.

phillipbarron (approved):
LGTM 👍

PR #145, opened 2023-05-22
Styling consistency tweaks

phillipbarron:
## What does this change?

Updates buttons in home menu to same size and the create new draft button now looks the same as the existing set.

## How to test

No functional changes introdused - if it looks right, it is.

It's all I do

## Images
Screenshot 2023-05-22 at 09 08 26

Screenshot 2023-05-22 at 09 07 05

### DIFF
--- a/apps/newsletters-ui/src/app/components/views/DraftListView.tsx
+++ b/apps/newsletters-ui/src/app/components/views/DraftListView.tsx
@@ -1,12 +1,11 @@
 import AddIcon from '@mui/icons-material/Add';
 import { Box, Button, Container } from '@mui/material';
-import { useLoaderData, useNavigate } from 'react-router-dom';
+import { useLoaderData } from 'react-router-dom';
 import { isDraft } from '@newsletters-nx/newsletters-data-client';
 import { DraftsTable } from '../DraftsTable';
 
 export const DraftListView = () => {
 	const list = useLoaderData();
-	const navigate = useNavigate();
 	if (!list || !Array.isArray(list)) {
 		return <nav>No Drafts</nav>;
 	}
@@ -25,7 +24,7 @@ export const DraftListView = () => {
 					<Button
 						variant="contained"
 						endIcon={<AddIcon />}
-						onClick={() => navigate('/newsletters/newsletter-data')}
+						href={'/newsletters/newsletter-data'}
 						aria-label="open create new draft wizard"
 					>
 						New draft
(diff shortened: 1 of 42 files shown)


apps/newsletters-ui/src/app/components/views/DraftListView.tsx
```diff
@@ -1,31 +1,35 @@
 import AddIcon from '@mui/icons-material/Add';
-import { Box, Container } from '@mui/material';
-import { useLoaderData } from 'react-router-dom';
+import { Box, Button, Container } from '@mui/material';
+import { useLoaderData, useNavigate } from 'react-router-dom';
 import { isDraft } from '@newsletters-nx/newsletters-data-client';
 import { DraftsTable } from '../DraftsTable';
-import { NavigateFab } from '../NavigateFab';
```
dblatcher, line 6: May as well delete `NavigateFab` if we aren't using it here - this is the only page it was on. It was basically a copy of the NavigateButton, but use `Fab` instead of `Button` as the base.

apps/newsletters-ui/src/app/components/views/DraftListView.tsx
```diff
@@ -1,31 +1,35 @@
...
 			<DraftsTable drafts={drafts} />
 			<Container maxWidth="lg">
-				<Box paddingX={1} display={'flex'} justifyContent={'flex-end'}>
-					<NavigateFab
-						href="/newsletters/newsletter-data"
-						color="primary"
-						variant="extended"
+				<Box
+					paddingX={1}
+					paddingBottom={'8px'}
```
dblatcher, line 21:
```suggestion
					paddingBottom={1}
```
using `1` here will give us the default MUI padding of 8px - I think the padding size is configurable.

apps/newsletters-ui/src/app/components/views/DraftListView.tsx
```diff
@@ -1,31 +1,35 @@
...
+					display={'flex'}
+					justifyContent={'flex-end'}
+				>
+					<Button
```
dblatcher, line 25: If you use the `NavigateButton` component with a `href` prop instead of `onClick`, the button will have the same appearance, but be rendered as a link component rather than a button (slightly more accessible), but still using the react-router `useNavigate` under the hood.

dblatcher (approved):
Looks good - couple of small suggestions.

dblatcher (approved):
Looks good - couple of small suggestions.

apps/newsletters-ui/src/app/components/views/DraftListView.tsx
phillipbarron: Ahh nice - top tip 👍

PR #146, opened 2023-05-22
Dblatcher no nx cache

dblatcher:
## What does this change?

Uninstalls the nx-cloud dependency and removes the configuration that uses it. Builds will use local caching, rather than remote caching.

We decided to do this after the [builds stopped working](https://github.com/guardian/newsletters-nx/actions/runs/5043176034/jobs/9044634561?pr=136) because we hadn't signed up for an NX account to claim the cache.

Able to run CI without managing an NX cloud account.

Builds will sometimes be slower locally - we can't use a cached build of a project built by and other developer and pushed to the cloud.

CI builds in github will be slower. Previously, PR's that only changed one project (EG the UI) did not need to build the depended projects (EG the data-client) for testing as they could take the compiled code from the remote cache (I think that was how it worked).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,26 +11,6 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  main:
-    name: Nx Cloud - Main Job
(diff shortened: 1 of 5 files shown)


phillipbarron (approved):
lgtm

PR #148, opened 2023-05-22
Dblatcher wizard style tweaks

dblatcher:
## What does this change?

 - replace some manual styling for the Wizard components with MUI components / props
 - simplify the custom styling for the `StepNav`
 - refactor the wizard buttons to a `WizardActionButton` component
 - refactor the api call in the `Wizard` component to use async/await

## How to test

Should work as before.

Easier to work with the Wizard component - now less busy.
less hard coded css value strings outside of the app-theme - easier to customise.

Did consider replacing the '.left-aligned-step-button' with [custom MUI overrides](https://mui.com/material-ui/customization/theme-components/#overrides-based-on-props) - although it seemed to be possible in theory, the resulting code would have been more obscure than using the 'global' styles in DefaultStyles module.


idrise (approved):
Nice improvements!

PR #154, opened 2023-05-24
Added alarms for unhealthy instance and > 5% 500 rate

phillipbarron:
## What does this change?

Adds alarms to both the Tool and API instances where:

1. The instances are unhealthy (ie, stop responding to health-checks)
2. We see an error rate (ie, return a 5xx response) at over 5% of our response rate

The errors are written to the existing newletter SNS which results in an email to the newsletters team.

## How to test

Tricky, actually. We need to cause some 5xx errors. Might be worth adding an endpoint that throws an error for testing and then verifying that we all get the email.

Edit:
build 677 (deployed in code at the time of writing) has a test endpoint added on `/api/boom` that will retun a 500 - hit that a few times and you should get an email

It email us when we get a. bunch of 500s

This might be noisy - we may need to tweak the error percentage such that we hear about serious incidents bit its not so noisy that we start to ignore it. 

We may also want to have different error rates across dev / prod. We can iterate and find the right level.

## Accessibility

### DIFF
--- a/cdk/lib/newsletters-tool.ts
+++ b/cdk/lib/newsletters-tool.ts
@@ -119,11 +119,16 @@ EOL`,
 			],
 		});
 
+		const alarmsTopicName = 'newsletters-alerts';
 		/** Sets up Node app to be run in EC2 */
 		const ec2AppTool = new GuNodeApp(this, {
 			access: { scope: AccessScope.PUBLIC },
 			certificateProps: { domainName: domainNameTool },
-			monitoringConfiguration: { noMonitoring: true },
+			monitoringConfiguration: {
+				http5xxAlarm: { tolerated5xxPercentage: 5 },
+				snsTopicName: alarmsTopicName,
+				unhealthyInstancesAlarm: true,
+			},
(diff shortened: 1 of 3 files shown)


cdk/lib/newsletters-tool.ts
```diff
@@ -146,7 +151,11 @@ EOL`,
 		const ec2AppApi = new GuNodeApp(this, {
 			access: { scope: AccessScope.PUBLIC },
 			certificateProps: { domainName: domainNameApi },
-			monitoringConfiguration: { noMonitoring: true },
+			monitoringConfiguration: {
+				http5xxAlarm: { tolerated5xxPercentage: 5 },
+				snsTopicName: alarmsTopicName,
+				unhealthyInstancesAlarm: true,
+			},
```
idrise, line 158: Not as complex as I expected.

idrise (approved):

cdk/lib/newsletters-tool.ts
phillipbarron: DevX have made our lives simpler here. 😄

PR #155, opened 2023-05-24
Rachel/tag setup validation

codingWithRach:
## What does this change?

https://github.com/guardian/newsletters-nx/pull/142  removed validation from the Tag Setup step.  In the case of the series tag it was correct to remove the validation (it should be done on the formSchema object if required).  However, the validation for the combination of composerTag and composerCampaignTag should remain i.e. if the user enters either one, then we need both

## How to test

`npm run dev`
Edit one of the draft newsletters and navigate to the Tag Setup step of the wizard
It's possible to press Next if neither the composer tag or composer campaign tag have been entered
However, if one or the other is entered, validation ensures that both have been entered

Field entries are correctly validated

## Images

Screenshot 2023-05-30 at 12 03 00

Screenshot 2023-05-30 at 12 03 26

## Accessibility

### DIFF
--- a/apps/newsletters-ui/src/app/components/NewsletterDataDetails.tsx
+++ b/apps/newsletters-ui/src/app/components/NewsletterDataDetails.tsx
@@ -99,7 +99,7 @@ export const NewsletterDataDetails = ({ newsletter }: Props) => {
 				<NavigateButton href="../" variant="outlined">
 					Back to List
 				</NavigateButton>
-				{/* TO DO - restrict the access to the JSON editter based on user role? */}
+				{/* TO DO - restrict the access to the JSON editor based on user role? */}
 				<RawDataDialog
 					record={newsletter}
 					title={newsletter.identityName}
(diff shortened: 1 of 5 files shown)


dblatcher:
This would work, but if we need either both or neither of these values to be set, that should ideally be reflected on the `NewsletterData` type 🤔 .  It sounds like we aren't modelling the values right - at the moment they are both optional strings, but maybe need to have them both in an optional nested object with two string fields.

It might be worth getting a product steer before doing too much work on that though - they are looking at changing how embed tags work.

dblatcher:
> This would work, but if we need either both or neither of these values to be set, that should ideally be reflected on the `NewsletterData` type 🤔 . It sounds like we aren't modelling the values right - at the moment they are both optional strings, but maybe need to have them both in an optional nested object with two string fields.
> 
> It might be worth getting a product steer before doing too much work on that though - they are looking at changing how embed tags work.

Meant to add - happy to approve this PR now unless you'd rather look at changing the schema?

dblatcher (approved):

PR #157, opened 2023-05-24
Dblatcher access profile in api

dblatcher:
## What does this change?

Extracts the code for getting the user profile form the `req.header` out of the `registerUserRoute` module

## How to test

(no changes to behaviour)

By importing the `getUserProfile` function into other parts of the API code, we will be able to use the user profile server-side, as well as client side via the existing  'whoami' end point.

Possible use cases:
 - code can include the user's email in the audit logs
 - we can restrict access to some API calls based on the user's identity (ideally, we will use a google user group to define permissions, but initially a list could be hard coded)
 - A future messaging service can send update/confirmation emails to the user that created a draft newsletter

We will need to be mindful of how we use the profile data.

### DIFF
--- a/apps/newsletters-api/src/app/get-user-profile.ts
+++ b/apps/newsletters-api/src/app/get-user-profile.ts
@@ -30,19 +30,22 @@ function parseJwt(
 
 export const getUserProfile = (
 	req: FastifyRequest,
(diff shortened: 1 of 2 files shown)


phillipbarron (approved):
Lovely

PR #158, opened 2023-05-25
Dblatcher remove unnecessary custom validation hooks

dblatcher:
## What does this change?

Removes custom validation hooks (`onBeforeStepChangeValidate`) which were only testing for missing values that the schema check now handle.

Makes the form schemas for the create draft wizard more restrictive than the `newsletterDataSchema` it is based on so that:
 - [to do]

and removes the custom validation for the related steps as the schema will now handle the validation.

## How to test

Wizards will work as before except:
-  the form elements will display validation warning (eg 'can't be empty') where previously, the error would only be flagged after submitting the form.
- we don't get custom validation errors, just the somewhat clunky ones parsed from the zod issues.
- The 'completion' status shown in the StepNav (based on the form schemas) will be more accurate

Less custom code to maintain!
Users get more feedback on editing rather than when they submit the step

It was nicer to be able to derive the form schemas by just 'picking' the fields - but the trade-off in complexity with the custom hooks is probably a benefit.

The error messages generated from the zod schema (eg 'Please try again: VALIDATION ERRORS: signUpDescription: "Must not be empty"') are not as clear as the custom error messages (eg "Please try again: NO HEADLINE PROVIDED").  We could address this with some better validation error handling in the state machine code (at present, validation messages are returned as strings rather than data see https://trello.com/c/eNRynPNu/418-change-return-type-for-executemodify).

### DIFF
--- a/apps/newsletters-ui/src/main.tsx
+++ b/apps/newsletters-ui/src/main.tsx
@@ -2,11 +2,11 @@ import { ThemeProvider } from '@mui/material';
 import { StrictMode } from 'react';
 import * as ReactDOM from 'react-dom/client';
 import { createBrowserRouter, RouterProvider } from 'react-router-dom';
+import { appTheme } from './app-theme';
 import { DefaultStyles } from './app/components/DefaultStyles';
 import { draftRoute } from './app/routes/drafts';
 import { homeRoute } from './app/routes/home';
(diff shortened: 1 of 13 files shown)


codingWithRach (approved):
LGTM

PR #164, opened 2023-06-01
Dblatcher wizard failure objects

dblatcher:
## What does this change?

The validation hooks (`AsyncValidator` and `Validator` types used for `onAfterStepStartValidate` and `onBeforeStepChangeValidate`) return an object with a message property and optional `FailureDetails` if the validation fails (instead of a string). The return value is still `undefined` if the validation passes.

The `AsyncExecution` and `Execution` functions (for `executeStep` and `executeSkip`) will return a `WizardExecutionFailure`(instead of a string - has the message property and same optional `FailureDetails` as the validation failures)  or `WizardExecutionSuccess`

Where the execution or validation fails, the message and the details are passed down through the API layer and UI.

The `FailureAlert` component uses the details(if present) as well as the message to provide more structured / readable feedback than it could with string error messages alone.

## How to test

Run the launch wizard for "avon-recruitment", but change the identityName to "blue-empowering" - the error details from the"check-input-is-unique" module will be include a list of "problems"  rendered as a list.

Try to submit any step with invalid input (eg empty string for newsletter's name) - the schema validation errors will be included in the error details and rendered nicely within the alert.

We are able to provide richer feedback on error states / user errors than we can with strings.

The `FailureDetails` type can be expanded to add extra metadata that we might want use for logging/analysis/monitoring 

A little more complexity, but should be offset by removing the confusion from interpreting strings as indicating an error.

## Images

Screenshot 2023-06-01 at 13 06 55

Screenshot 2023-06-01 at 14 42 54

### DIFF
--- a/apps/newsletters-ui/src/app/components/Wizard.tsx
+++ b/apps/newsletters-ui/src/app/components/Wizard.tsx
@@ -17,6 +17,7 @@ import { MarkdownView } from './MarkdownView';
 import { StateEditForm } from './StateEditForm';
 import { StepNav } from './StepNav';
 import { WizardActionButton } from './WizardActionButton';
+import { ZodIssuesReport } from './ZodIssuesReport';
 
 /**
  * Interface for the props passed to the `Wizard` component.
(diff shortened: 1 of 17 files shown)


phillipbarron (approved):
Useful change 👍

PR #165, opened 2023-06-01
Dblatcher radio enum fields

dblatcher:
## What does this change?

The `SchemaForm` will render a set of radio buttons instead of a drop down, unless the number of options exceeds the new `maxOptionsForRadioButtons` prop.

This prop is set to 5 for all existing uses of the SchemaForm`.

## How to test
Open a form page eg 
http://localhost:4200/newsletters/edit/blue-empowering

or go through a wizard to see some radio buttons.

Should be safe. The change is only presentational.

## Images

Screenshot 2023-06-01 at 17 32 51


phillipbarron (approved):
LGTM

PR #166, opened 2023-06-02
use semantic elements and aria labels

dblatcher:
## What does this change?

Puts some semantic mark-up and aria-labels on the step nav:
 - the Stepper itself uses a 'nav', grouping the Steps together on the accessibility tree
 - the Steps not containing buttons use 'section', to group and label the contents
 - the Steps containing buttons remain as 'div' (no semantic significance), as the 'button' within it is the semantic element

## How to test

view the accessibility tree on a wizard page - the document structure is clearer.

The wizard pages would be more accessible to screen reader users

This is a small improvement, but would need proper user testing to asses how intuitive the pages are.

## Accessibility

|before|after|
|---|---|
| Screenshot 2023-06-02 at 08 56 04 | Screenshot 2023-06-02 at 09 26 42 |

### DIFF
--- a/apps/newsletters-ui/src/app/components/StepNav.tsx
+++ b/apps/newsletters-ui/src/app/components/StepNav.tsx
@@ -56,6 +56,30 @@ const CompletionCaption = (props: { status: StepStatus | undefined }) => {
 	}
 };
 
+const ariaLabelForNonButtonStep = (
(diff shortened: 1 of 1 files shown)


phillipbarron (approved):
LGTM

PR #167, opened 2023-06-02
Dblatcher limit edit access

dblatcher:
## What does this change?

Adds a function to up the user profile against a static list of users and retrieve a set of `UserPermissions` based on the `UserAccessLevel` for the user.

The API routes for editing Newsletters and handling wizard requests use this function and return a 403 error if the user lacks the required permission.

## How to test

To test locally, you will need to make the existing `getUserProfile` function provide a user object, despite not having the header set by the google auth. You can do this by:
 - set the FAKE_JWT and USE_FAKE_JWT=true local environment variable to give yourself a 'fake' user profile locally. You can get the value for the FAKE_JWT by accessing the request headers from CODE (probably more trouble than its worth); or
 - hard code your user profile and change `getUserProfile` to return it by default

When you have your local user profile (hover over the avatar icon in the UI to check the name is coming through), add the email address to the `STATIC_USERS` object in the user-permissions module, assigning the desired `UserAccessLevel`.

If you set the level to "editor" or "developer" the UI will work as before - if not, you will get user-facing error messages when trying to run the launch wizard or use the 'edit newsletter' pages.

Hard coding the users is an interim solution / WIP .It should be possible to replace the static users with an API call to the google groups service (or similar).

It would have been possible to provide the user permissions to the `launchService` interface and have the Wizard request a markdown message explaining that the particular user does not have permission, but that seemed excessively complex and potentially fragile - it seemed better to block the request at the API level and avoid complicating the existing wizard logic further.

We do need to be wary of 'doing auth' within our application. Provided we are clear on which routes need user-level access logic, it should be okay.

Our error responses  are currently limited to strings - by using the changes in https://github.com/guardian/newsletters-nx/pull/164 we could provide more structured details about the error that the UI could use to render more helpful feedback.

## Images

Screenshot 2023-06-02 at 12 16 36

### DIFF
--- a/apps/newsletters-api/src/app/routes/currentStep.ts
+++ b/apps/newsletters-api/src/app/routes/currentStep.ts
@@ -1,5 +1,7 @@
 import type { FastifyInstance } from 'fastify';
 import { newslettersWorkflowStepLayout } from '@newsletters-nx/newsletter-workflow';
+import type { UserProfile } from '@newsletters-nx/newsletters-data-client';
+import { getPermissions } from '@newsletters-nx/newsletters-data-client';
 import type {
 	CurrentStepRouteRequest,
 	CurrentStepRouteResponse,
@@ -10,6 +12,7 @@ import {
 	StateMachineErrorCode,
 } from '@newsletters-nx/state-machine';
 import { draftStore, launchService } from '../../services/storage';
+import { getUserProfile } from '../get-user-profile';
 
 const getHttpCode = (error: StateMachineError): number => {
 	switch (error.code) {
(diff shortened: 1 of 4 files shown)


phillipbarron (commented):
Tested locally and working as expected. 
Agree that a better UX would be not to present options that are not available to the user. In this version, we allow them to edit and then prevent update. Ideally, users with `readOnly` access should not be able to:

- Create Drafts
- Edit Drafts
- Edit Raw Data
- Access Edit Routes

This can be iterated upon but in the short term:


We should position the error messages inside the content wrapper

Screenshot 2023-06-02 at 15 04 11

Let's ignore the hiding controls for `readOnly` for now

This is great btw - you write code as fast as I think of it

dblatcher:
> Tested locally and working as expected. Agree that a better UX would be not to present options that are not available to the user. In this version, we allow them to edit and then prevent update. Ideally, users with `readOnly` access should not be able to:
> 
> * Create Drafts
> * Edit Drafts
> * Edit Raw Data
> * Access Edit Routes
> 

Sounds good - I'll make the default 'viewer' access level unable to create and edit drafts.
UX wise, we might want to show the options in a 'disabled' state rather than hide the buttons altogether - that the user know that there is a feature there but they can't use it - rather than thinking they just can't find the button.

phillipbarron (approved):
tiny sugestion around naming

PR #172, opened 2023-06-07
Pb/date inputs and copy tweaks

phillipbarron:
## What does this change?

Uses the MUI DatePicker Component in place of the text input. This resolves an issue where both label & format were rendered in the date inputs & gives us consistent look & feel across browsers.

Also contains some updates to copy as requestsed in [the review ticket](https://trello.com/c/9DG4mVyW/477-review-newsletter-tool-for-feedback-ahead-of-editorial-demo)

## How to test

Check it out, set some dates, fail to set some dates. 

We can manage dates in the same way as we could prior to the change

Think so

## Images

Screenshot 2023-06-07 at 07 59 35

Screenshot 2023-06-07 at 07 59 17


dblatcher (approved):
Looks great! 
Its not needed now (and wasn't supported before), but it would be nice to have the `DateInput` component handle the "optional" prop in case this comes up.

PR #173, opened 2023-06-07
store newsletter metadata

dblatcher:
## What does this change?

Newsletters (not drafts) in storage have a metadata property recording which user created them (and when) and when and by whom they were last modified.

The Newsletters storage classes  methods will 
 - add a new metadata object in the 'create' method
 - update the existing metadata in the 'update' method (in s3, if the metadata is absent from the file, a new metadata will be added with the `createdBy` as "unknown" and `createdTimestamp` as 0, but the last modified info added)
 - remove the metadata before returning the newsletter in the 'read' methods (except in the new 'readWithMeta' methods - see below)

https://trello.com/c/V6rSv3kW/465-model-audit-data-in-newsletter-data

## How to test

Should work as before, but the data in s3 or the in memory storage will include meta data, that is not exposed to the UI / API responses.

There is a new newsletterStorage method and API route for exposing the metadata, but it requires 'developer' access
http://localhost:3000/api/newsletters/meta/blue-empowering
This is useful for testing, but we can remove the route if don't ever want the meta data to be available on the API / to the application. 

Able to track which users made which changes with the combination of this metadata and the s3 versioning.

Need to decide how visible the meta data should be - should we remove the metadata route?


dblatcher:
TO DO - test on code

phillipbarron (commented):
Tested locally using s3 based storage (using the `npm run dev:s3` option with a bucket set up in devPlayground)

Works as described 🎉 

Screenshot 2023-06-07 at 20 54 18

Note the different email addresses and timestamps.

I had imagined that we would keep this metadata for drafts too. If someone has spend a bunch on timme working on a ddraft and those changes are overwritten, it's nice to know who and when.

We should also update the migration scripts to use the creationDate from the exising doc and to write that in to the new metadata object

We are mixing tense a bit in the metadata description - this is a (my) mistake in the ticket - not yours:

Would be good to rename `creationTimestamp` to `createdTimestamp` then both the verbs are past tense whech reads more clearly.

dblatcher:
> 
> We are mixing tense a bit in the metadata description - this is a (my) mistake in the ticket - not yours:
> 
> Would be good to rename `creationTimestamp` to `createdTimestamp` then both the verbs are past tense whech reads more clearly.

👍  made that change

dblatcher:
> I had imagined that we would keep this metadata for drafts too. If someone has spend a bunch on time working on a draft and those changes are overwritten, it's nice to know who and when.
> 
> We should also update the migration scripts to use the creationDate from the exising doc and to write that in to the new metadata object
> 

Agreed - but I'd say that should be a separate PR. 

It will be a pretty large diff to provide the user profile to the service used by the draft wizard and update all the steps.

phillipbarron:
> > I had imagined that we would keep this metadata for drafts too. If someone has spend a bunch on time working on a draft and those changes are overwritten, it's nice to know who and when.
> > We should also update the migration scripts to use the creationDate from the exising doc and to write that in to the new metadata object
> 
> Agreed - but I'd say that should be a separate PR.
> 
> It will be a pretty large diff to provide the user profile to the service used by the draft wizard and update all the steps.

Sure - sounds sensible - I will create a ticket in Trello-land to add in the metadata for the Draft content.

PR #175, opened 2023-06-07
Dblatcher cant complete by visiting

dblatcher:
## What does this change?

https://trello.com/c/oxA0mb80/482-bug-fields-marked-as-complete-when-vistied-in-the-wizard

Visiting the 'multi thrasher' step will no longer cause it to be marked 'complete'. This was happening because visiting the step caused the field to be populated with an empty array. Fixed by making the logic in "resolve-step-status" consider a field in an optional step be 'unset' if it is false-y or an empty array (was only checking false-iness before).

"Single thrasher" & "Newsletter Design" start marked as 'optional' and will be marked 'complete' only if any of the fields are set to a non-empty and non-false-y value. This required two changes:
 - steps are considered "optional" if the initial/default values for schema passed schema validation
 - the newsletter data starts with default values populated (string values to empty string, booleans to false) - these default s would previously only be set when the user visits the step (which was is was 'required' to visit them, even though the initial defaults were valid) 

Users get a clearer picture of what steps they can ignore. Data doesn't change by visiting a step and taking no action.

Users can now ignore 'optional' steps by skipping past them, without considering if the default values are actually what they want.


phillipbarron (commented):
Checked out locally - see this is draft & so apologies for early comment but...
After this change, it's not possible to unset values in the design layout.

Try starting the wizard, add a name and then add figma info. Click next, navigate back & try to remove.

phillipbarron (approved):
LGTM - Tested locally, behaves as expected

PR #177, opened 2023-06-08
Dblatcher wizard feedback

dblatcher:
## What does this change?
Makes some changes to the UI in response to feedback:

- Remove legacy-article based from options presented to user in the wizard step
- Show message about rendering options wizard at the last step of the data wizard (if the draft is article-based)
- Add margin between menu and header (h2 on the theme)
- Add page headings to View launched newsletters and view drafts pages 
- remove the 'restricted' column from the newsletters table

https://trello.com/c/byCZHVIi/483-team-demo-feedback


phillipbarron (approved):
LGTM

PR #179, opened 2023-06-08
Dblatcher skipping modal

dblatcher:
## What does this change?

The `executeSkip` function used in the wizardLayouts will send the version of the data previously persisted  on the server back to the client - effectively causing the client to 'discard' the local changes to the data the user may have made before skipping from the current step.

Adds an optional boolean property to the `WizardLayoutStep` to specify that skipping **FROM** that step will persist the data (the default behaviour is  assumed to be that skipping will NOT persist the local changes). This is set to `true` on the `createDraftNewsletterLayout` since it uses `executeCreate` rather than `executeSkip`.

Before the user skips and discard the local changes on the step, they are presented with a confirmation modal. The modal is NOT presented if the user has not made any changes on the step since they landed on it.

## How to test

Run the wizard - go to any step other than the initial `createDraftNewsletterLayout`. Make a change to a field and click another step to skip. The modal will offer the choice of staying on the current step or discarding your local change.

Clearer to users what is saved and what is not. No mismatch between the client and server leading to unexpected behaviour.

yes

## Images

Screenshot 2023-06-08 at 15 26 37

### DIFF
--- a/apps/newsletters-ui/src/app/components/SkipConfirmationDialog.tsx
+++ b/apps/newsletters-ui/src/app/components/SkipConfirmationDialog.tsx
@@ -1,0 +1,35 @@
+import {
+	Button,
+	Dialog,
+	DialogActions,
+	DialogContent,
+	DialogTitle,
+} from '@mui/material';
+
+interface Props {
+	currentStepId: string;
+	showSkipModalFor?: string;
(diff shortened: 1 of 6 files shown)


phillipbarron (approved):
Looks great

PR #183, opened 2023-06-12
Hide migration creation date in table

phillipbarron:
## What does this change?

We are, arbitrarily, using 946684800 as the migration timestamp in the [migration scripts](https://github.com/guardian/newsletters-migration-tool). This change checks for that date in the table date formatting and returns unknown instead.

## How to test

Check it out, manually update some newsletter to have the creation date of `946684800` (a number, not a string). Now review in the launched newsletters table. It should state `Unknown` rather than Sunday aa Jan, 1970 or whatever it was. 

Screenshot 2023-06-12 at 14 13 19

### DIFF
--- a/apps/newsletters-ui/src/app/components/Cell.tsx
+++ b/apps/newsletters-ui/src/app/components/Cell.tsx
@@ -1,13 +1,14 @@
 import { renderYesNo } from '../util';
 
 export type Cell<T> = { cell: { value: T } };
-
+export const MIGRATION_TIMESTAMP_VALUE = 946684800;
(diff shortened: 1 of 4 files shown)


apps/newsletters-ui/src/app/components/Cell.tsx
```diff
@@ -1,13 +1,14 @@
 import { renderYesNo } from '../util';
 
 export type Cell<T> = { cell: { value: T } };
-
+export const MIGRATION_TIMESTAMP_VALUE = 946684800;
```
dblatcher, line 4: We should probably define this value in the data-client library and import it from there in case we need it elsewhere

dblatcher (approved):

PR #184, opened 2023-06-12
Dblatcher fix sending undefined

dblatcher:
## What does this change?

 - When the Wizard UI sends a value explicitly set to  `undefined`, the `JSON.stringify` call is configured to replace each `undefined` with a `null` (keys for `undefined` values are ignored, since `undefined` is not valid JSON, but `null` is).
 
 - The `handleWizardRequestAndReturnWizardResponse`  uses `replaceNullWithUndefined` to replace `null`s in the formData with `undefined` before acting on the data.

 - The `formDataToDraftNewsletterData` function does not ignore `undefined` values

Net result - user can send a value of `undefined` to replace an existing truthey value with `undefined` on the server.

## How to test

 - `npm run dev`
 - go tp http://localhost:4200/newsletters/newsletter-data/5770 to open the editor for "Avon Recumbent"
 - skip to the "newsletter design" step, empty the value for "design brief doc"
 - hit "back" to submit the value, bypassing custom validation (which you probably shouldn't be be able to do, but works for the purpose of this fix)
- go to http://localhost:4200/drafts/5770 to see the data - "design brief doc" will be gone

Happier users, no manual editing of the JSON document in s3.

might have similar bugs elsewhere? should we do the same with other stringified data with meaningful `undefined` values?

### DIFF
--- a/apps/newsletters-api/src/services/storage/inMemoryStorageInstance.ts
+++ b/apps/newsletters-api/src/services/storage/inMemoryStorageInstance.ts
@@ -26,6 +26,8 @@ export const makeInMemoryStorageInstance = () =>
 			thrasherDate: new Date('1970-01-02T00:21:18.876Z'),
 			privateUntilLaunch: false,
 			onlineArticle: 'Web for all sends',
+			designBriefDoc: 'Its on google drive somewhere.',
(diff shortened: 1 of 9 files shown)


phillipbarron (approved):
All looks very sensible

PR #185, opened 2023-06-16
accept hyphens and underscores in identiytname as what's on uses this

phillipbarron:
## What does this change?

Adds a new schema helper which permits a mix of underscore and hyphens to support the existing formats. The help is used in `identityName` 
Update - to support existing values, have converted the existing `underscoreCasedString` has been update to be alpha-numeric

## How to test

Verify that `what's on`, `Tokyo2020` and `guardian_weekly_newsletter` are returned 

What's on is back on

this will permit a variation in formats, but it's already possible. We might wish to lock this down later / migrate existing data, but that's a task for another day,

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This is a monorepo for:
 
 -   defining the data model for editorial newsletters at the Guardian
 -   serving an API to access the newsletter data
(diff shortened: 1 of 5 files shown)


dblatcher (approved):
👍

PR #188, opened 2023-06-20
Dblatcher don't render forbidden buttons

dblatcher:
## What does this change?

Updates the UI to ~~disable~~ **not render** the buttons that navigate to pages the user will lack the permissions to use. EG the buttons linking to the wizard pages.

~~The disabled buttons have tool tips with a message explaining why they are disabled.~~

## How to test

Update your .env.local file to give yourself "viewer" permissions. 

Users are only presented with the UI for the actions they are able to do.

This is only a frontend change - the enforcement of the permissions was already in place on the API.


apps/newsletters-ui/src/app/components/RawDataDialog.tsx
```diff
@@ -78,7 +80,15 @@ export const RawDataDialog = ({
 				<DialogActions>
 					<Button onClick={copyJson}>copy json</Button>
 					{editHref && (
-						<NavigateButton href={editHref}>edit json</NavigateButton>
+						<NavigateButton
+							href={editHref}
+							disabled={!useJsonEditor}
+							toolTip={
+								!useJsonEditor ? "You don't have access to that" : undefined
```
codingWithRach, line 87: Might be better as "You don't have access to the JSON editor" ??

codingWithRach (approved):
Looks good.  Back in the day, I was always told it was better to hide things that people didn't have access to, but that might be another question for UX

phillipbarron (approved):
I will try to dig out the reference for this but I understand, in general, it's better UX only to present usable controls. So rather than a disabled button, perhaps for users without exit / launch permissions, we should omit the columns from the table?

I think that this is an improvement so please go ahead and merge but perhaps we could hide stuff we can't use as an enhancement?

phillipbarron:
Discussed this with Abo today - His view is the same regarding hiding. In general, it's a good design practice not to present unusable controls. We do not, for example, present any disabled controls in the front end. Whilst there is no convention for internal tooling, I'd say that hiding is in the way to go.

dblatcher:
> Discussed this with Abo today - His view is the same regarding hiding. In general, it's a good design practice not to present unusable controls. We do not, for example, present any disabled controls in the front end. Whilst there is no convention for internal tooling, I'd say that hiding is in the way to go.

I'll mark this PR as a draft and re-purpose it to hide the buttons.

libs/newsletters-data-client/src/lib/user-profile.ts
```diff
@@ -56,5 +57,23 @@ export const levelToPermissions = (
 			UserAccessLevel.Drafter,
 		].includes(accessLevel),
 		viewMetaData: [UserAccessLevel.Developer].includes(accessLevel),
+		useJsonEditor: [UserAccessLevel.Developer].includes(accessLevel),
 	};
 };
+
+export const noPermissionMessage = (
```
codingWithRach, line 64: Not sure where this is called?  (I did a search but couldn't find anything, but it's always possible I did something silly)

codingWithRach (commented):
Works well.  There might be an argument for changing the copy on this page, depending on whether a person has permissions to access the launch wizard:
Screenshot 2023-06-28 at 12 25 45

libs/newsletters-data-client/src/lib/user-profile.ts
dblatcher: oops - I removed that, but didn't push the change!

codingWithRach:
> > Works well. There might be an argument for changing the copy on this page, depending on whether a person has permissions to access the launch wizard
> 
> Good point!
> 
> Currently, the `dynamicMarkdown` hooks on the Steps don't have access the the Wizard's `serviceInterface`, which it would need in order to get the permissions for the user stored on the instance of `DraftService`. It would take some plumbing to make that possible, but it would be a great feature.
> 
> Interesting point - with access to the `serviceInterface` and the `userProfile`, we could also include the user's details in the markdown messages , eg "I'm sorry, I can't do that, {{userName}}" - but that might be a bit unnecessary.
> 
> For now, I could just change the copy to something like:
> 
> ```
> Once all the data required to create **{{name}}** has been specified, [the launch wizard](/newsletters/launch-newsletter/{{listId}}) can be used to launch your new newsletter. If you don't have access to the launch, please contact the newsletters team.
> ```

Sounds a sensible approach.  It might be worth doing the plumbing at some point in future, or it might be deemed not to provide sufficient benefit (in comparison to features such as automation)

PR #191, opened 2023-06-26
Dbltacher rendering option ready bug

dblatcher:
## What does this change?

Fixes some logic gaps and aligns the validation to the UI:
 - `renderingOptions` must be set on an  "article-rendering" draft 
 - the boolean values in `renderingOptions` (not optional in the schema) not expliclity set are defaulted from `undefined` to `false` when launching or checking readiness.

## How to test

on main
 - no issues would be raised on the details page for an "article-rendering" draft with undefined renderingOptions
 - an "article-rendering" draft with a renderingOptions with boolean values set to `undefined` (IE from the "Optional" steps were skipped) would fail validation by the launch wizard and not be launched

on branch:
 - on the details page for an "article-rendering" draft with undefined `renderingOptions` the "rendering options is undefined" issue will be flagged
 - when launching and checking for issues,  undefined boolean values in the rendering options will default to false.

Less confused users.

Data input to store still needs to be valid, so should be safe.

### DIFF
--- a/libs/newsletter-workflow/src/lib/getInitialStateForLaunch.ts
+++ b/libs/newsletter-workflow/src/lib/getInitialStateForLaunch.ts
@@ -7,6 +7,7 @@ import type {
 } from '@newsletters-nx/newsletters-data-client';
 import {
 	addSuffixToMakeTokenUnique,
+	defaultRenderingOptionsValues,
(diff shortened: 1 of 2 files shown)


phillipbarron (approved):
LGTM

PR #194, opened 2023-06-30
remove line from markdown about automatic emails for tags

dblatcher:
## What does this change?

https://trello.com/c/ZL6Y2oBC/508-copy-for-tag-setup-page
Remove line from markdown about automatic emails for tags


phillipbarron (approved):
👍

PR #198, opened 2023-07-05
Cancelled newsletter API support

phillipbarron:
## What does this change?

The existing newsletters API can optionally include cancelled newsletters in the response based on a query param value. This PR adds that same fuctionality to the new API

## How to test

- Check out branch
- Run App
- Hit the [legacy endpoint](http://localhost:4200/api/legacy/newsletters) without a query param - should see no cancelled newsletters
- Hit the [legacy endpoint](http://localhost:4200/api/legacy/newsletters?includeCancelled=true) with a query param `includeCancelled=true`- Now they should appear
- Same for the non legacy API:
- [Without query param](http://localhost:4200/api/newsletters)
- [And with](http://localhost:4200/api/newsletters?includeCancelled=true)

It works

Think this is OK - Have hard code true to the UI call as we alway want to be able to manage from the tool

### DIFF
--- a/apps/newsletters-api/src/app/routes/newsletters.ts
+++ b/apps/newsletters-api/src/app/routes/newsletters.ts
@@ -12,30 +12,45 @@ import {
 	mapStorageFailureReasonToStatusCode,
 } from '../responses';
 
+interface IListNewslettersQuerystring {
+	includeCancelled?: string;
+}
 export function registerNewsletterRoutes(app: FastifyInstance) {
 	// not using the makeSuccess function on this route as
 	// we are emulating the response of the legacy API
-	app.get('/api/legacy/newsletters', async (req, res) => {
-		const storageResponse = await newsletterStore.list();
-		if (!storageResponse.ok) {
-			return res
-				.status(mapStorageFailureReasonToStatusCode(storageResponse.reason))
-				.send(makeErrorResponse(storageResponse.message));
-		}
+	app.get<{ Querystring: IListNewslettersQuerystring }>(
+		'/api/legacy/newsletters',
+		async (req, res) => {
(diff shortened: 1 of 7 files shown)


apps/newsletters-api/src/app/routes/newsletters.ts
```diff
@@ -12,30 +12,45 @@ import {
...
 	// not using the makeSuccess function on this route as
 	// we are emulating the response of the legacy API
-	app.get('/api/legacy/newsletters', async (req, res) => {
-		const storageResponse = await newsletterStore.list();
-		if (!storageResponse.ok) {
-			return res
-				.status(mapStorageFailureReasonToStatusCode(storageResponse.reason))
-				.send(makeErrorResponse(storageResponse.message));
-		}
+	app.get<{ Querystring: IListNewslettersQuerystring }>(
```
dblatcher, line 21: The legacy route should behave the same was as the current API endpoint (to avoid migration panic),   so we should not add the `includeCancelled` param on the legacy route but always include the cancelled newsletters by default.

phillipbarron: The intent here was to facilitate the param passing from the existing api.

It looks like it used to do this but doesn't anymore and I was thrown by the old config in fastly:
Screenshot 2023-07-05 at 15 04 30

That is only actually set on the code environment and we are not writing a `newsletters-including-cancelled` file to s3. There is one in the dev bucket, so it's eith something that was thought about and never released or partially deleted. I will add a ticket to remove the VCL snipped from Fastly to save confusion in the future. 

The short version of this is, we don't need this change. Closing the PR.

Thanks for looking

PR #199, opened 2023-07-05
Edit rendering options form

dblatcher:
## What does this change?

 - add a page for setting or editing the `renderingOptions`, and the `SeriesTag` and `Category` on a launched newsletter.
 - adapts the undefined -> null -> undefined mechanism used by the state machine to allow defined values on a newsletters to be changed to `undefined` with a json post to the API (`undefined` is not valid data, so key-value pairs with a value of `undefined` are omitted by JSON.stringify)

## How to test

The form pages  can be access from the "set rendering options" button on the homepage.

Setting series tag to an empty string and submitting will 'unset' the value on the server to undefined.

Launched newsletters can be "relaunched" article-based by adding the rendering options.
Editorial could use the form to adjust the options for an existing newsletter design.

Maybe we should 'hide' this page until email rendering is using the API to render emails?

## Images

Screenshot 2023-07-05 at 15 37 46

### DIFF
--- a/libs/newsletters-data-client/src/lib/json-undefined-null-conversions.spec.ts
+++ b/libs/newsletters-data-client/src/lib/json-undefined-null-conversions.spec.ts
@@ -1,4 +1,7 @@
-import { replaceNullWithUndefined } from './json-undefined-null-conversions';
+import {
+	replaceNullWithUndefinedForUnknown,
+	replaceNullWithUndefinedInFormDataRecord,
+} from './json-undefined-null-conversions';
 import type { FormDataRecord } from './transformWizardData';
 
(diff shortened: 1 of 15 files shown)


libs/newsletters-data-client/src/lib/json-undefined-null-conversions.spec.ts
```diff
@@ -0,0 +1,130 @@
+import {
```
phillipbarron, line 1: love a thorough test 🎉

phillipbarron (approved):
Looks great - get's us a long way toward relaunch 👍

PR #201, opened 2023-07-11
Dblatcher text areas in forms

dblatcher:
## What does this change?

Adds an option to have string values in a `WizardStepLayout` collected in multi-line text areas, so better support longer values like description fields, using an optional `fieldDisplayOptions` property.

The input on text areas is processed to remove any tabs or carriage returns - in effect the user cannot input formatted / multiline text, but the text will wrap, rather than be clipped. (We could add extra options to configure textArea to allow formatted text, but there is not a current need for this).

configures the steps in the main draft wizard to use the textArea for  text  fields likely to be longer (eg headlines, descriptions.

Makes the width for fields in schema forms wider - the wrapper was set to maxWidth = 400, but changed it to 'sm' (600px by default, but can be configured in the theme).

## How to test

Description fields will be multi-line, but not accept CR's or Tabs.

Easier for user to read longer text values.

One extra thing to remember when configuring new wizards. 

## Images

Screenshot 2023-07-11 at 12 05 09

### DIFF
--- a/apps/newsletters-ui/src/app/components/EditNewsletterForm.tsx
+++ b/apps/newsletters-ui/src/app/components/EditNewsletterForm.tsx
@@ -74,7 +74,7 @@ export const EditNewsletterForm = ({ originalItem }: Props) => {
 				}
 				stringConfig={{
 					signUpDescription: {
-						type: 'textArea',
+						inputType: 'textArea',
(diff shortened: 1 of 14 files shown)


phillipbarron (approved):
LGTM

PR #209, opened 2023-07-14
Dblatcher/is on code indicator

dblatcher:
## What does this change?

 Adds a 'Code' icon with a tooltip if the location hostName includes ".code"

## How to test

deploy to code.

Users feel able to experiment on CODE and don't get PROD and CODE mixed up

Using the hostname is easy, but in theory the project could be moved to a different set of hosts - however very unlikely the production version woudl be on a domain with ".code." in the name.

## Images (how would look on code)
Screenshot 2023-07-14 at 15 04 42


phillipbarron (approved):
Love it

PR #215, opened 2023-07-17
Added replace function for JSON edit to allow removal of properties

phillipbarron:
## What does this change?

Adds a POST endpoint which allows us to replace the content of a launched Newletter. This means we can remove properties (something not currently possible in the JSON editor as we merge edxisting state)

## How to test

Check it out, removed a property from a launched newsletter (`exampleUrl` is an option) , now update and verify the value is removed and that the item is still available from the API.

Should be Ok though we might want to consider access to this element of the app

### DIFF
--- a/apps/newsletters-api/src/app/routes/newsletters.ts
+++ b/apps/newsletters-api/src/app/routes/newsletters.ts
@@ -1,6 +1,13 @@
 import type { FastifyInstance } from 'fastify';
+import type {
+	NewsletterData,
+	NewsletterDataWithoutMeta,
+	SuccessfulStorageResponse,
+	UnsuccessfulStorageResponse,
+} from '@newsletters-nx/newsletters-data-client';
 import {
-	isPartialNewsletterData,
+	isNewsletterData,
+	isPartialNewsletterData, NewsletterStorage,
 	replaceNullWithUndefinedForUnknown,
 	transformDataToLegacyNewsletter,
 } from '@newsletters-nx/newsletters-data-client';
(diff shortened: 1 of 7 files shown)


apps/newsletters-api/src/app/routes/newsletters.ts
```diff
@@ -105,4 +107,51 @@ export function registerReadWriteNewsletterRoutes(app: FastifyInstance) {
...
+		if (isNaN(newsletterIdAsNumber)) {
+			return res.status(400).send(makeErrorResponse(`Non numeric id provided`));
+		}
+
+		replaceNullWithUndefinedForUnknown(newsletterData);
+
+		if (!isNewsletterData(newsletterData)) {
+			return res.status(400).send(makeErrorResponse(`Not a valid newsletter`));
+		}
+
```
dblatcher, line 138: I think it would be better to re-order this so we avoid unsafely casting `newsletter as NewsletterData` (I know its only a few lines before it gets validated) - if we do `replaceNukkWithUndefinedForUnknown`, then validate with `isNewsletterData`, `newsletter` will be type-safe

libs/newsletters-data-client/src/lib/newsletter-storage/InMemoryNewsletterStorage.ts
```diff
@@ -154,6 +154,28 @@ export class InMemoryNewsletterStorage implements NewsletterStorage {
...
+	async replace(
+		listId: number,
+		newsletter: NewsletterDataWithoutMeta,
+		user: UserProfile,
+	) {
+		const match = this.memory.find((item) => item.listId === listId);
+		if (!match) {
+			return Promise.resolve(this.buildNoItemError(listId));
+		}
+
```
dblatcher, line 166: There is a case for throwing an error if `newsletter.identityName !== match.identityName || newsletter.listId !== match.listId`.

libs/newsletters-data-client/src/lib/newsletter-storage/s3-newsletter-storage.ts
```diff
@@ -282,6 +281,46 @@ export class S3NewsletterStorage implements NewsletterStorage {
...
+	> {
+		const newsletterToUpdate = await this.fetchNewsletter(listId);
+		if (!newsletterToUpdate) {
+			return {
+				ok: false,
+				message: `failed to read newsletter with id ${listId}`,
+			};
+		}
+		const { identityName } = newsletterToUpdate;
+		const updatedNewsletter: NewsletterDataWithMeta = {
```
dblatcher, line 300: As with the in memory version - maybe throw an error if `newsletter.identityName !== newsletterToUpdate.identityName || newsletter.listId !== newsletterToUpdate.listId`.

apps/newsletters-api/src/app/routes/newsletters.ts
```diff
@@ -105,4 +107,51 @@ export function registerReadWriteNewsletterRoutes(app: FastifyInstance) {
...
 	});
+
+	app.post<{
+		Params: { newsletterId: string };
+		Body: unknown;
+	}>('/api/newsletters/:newsletterId', async (req, res) => {
+		const user = getUserProfile(req); //Todo: create middleware to get user profile
+		const accessDeniedError = await makeAccessDeniedApiResponse(
+			user.profile,
+			'editNewsletters',
```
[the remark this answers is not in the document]
phillipbarron: Agree - will add a ticket for this

dblatcher (approved):

PR #217, opened 2023-07-18
Dblatcher/render requests

dblatcher:
## What does this change?

 - Adds an environment variable to allow using a local version of email-rendering instead of PROD
 - new API routes to fetch demo content from email-rendering by posting `newsletterData` to it
- adds preview pages to show the demo content for a launched newsletter
- puts the preview alongside the 'edit rendering option' form so the user can see the effect of the changes

## How to test

try the new pages.

Uses better able to understand how the changes will effect the design.

Should do some user acceptance testing on CODE before merging? 

## Images

Screenshot 2023-07-19 at 12 54 46

### DIFF
--- a/apps/newsletters-api/src/app/routes/rendering-templates.ts
+++ b/apps/newsletters-api/src/app/routes/rendering-templates.ts
@@ -1,11 +1,19 @@
 import type { FastifyInstance } from 'fastify';
-import { makeErrorResponse, makeSuccessResponse } from '../responses';
+import { newsletterStore } from '../../services/storage';
+import {
+	makeErrorResponse,
+	makeSuccessResponse,
+	mapStorageFailureReasonToStatusCode,
+} from '../responses';
 
 export type RenderingTemplate = {
 	id: string;
 	status: string;
 	title: string;
 };
+
+const NEWSLETTER_RENDER_URL =
+	'http://localhost:3010/data-article/render-template';
 
 const TEMPLATES_LIST_URL =
 	'https://email-rendering.guardianapis.com/info/templates/';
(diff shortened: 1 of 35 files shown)


apps/newsletters-api/src/app/routes/rendering-templates.ts
```diff
@@ -1,14 +1,21 @@
...
 export type RenderingTemplate = {
 	id: string;
 	status: string;
 	title: string;
 };
 
-const TEMPLATES_LIST_URL =
-	'https://email-rendering.guardianapis.com/info/templates/';
+const NEWSLETTER_RENDER_URL = `${getEmailRenderingHost()}/data-article/render-template`;
+const TEMPLATES_LIST_URL = `${getEmailRenderingHost()}/info/templates/`;
```
phillipbarron, line 19: We could avoid evaluating that twice with an assign

```suggestion
const EMAIL_RENDERING_HOST = getEmailRenderingHost();
const NEWSLETTER_RENDER_URL = `${EMAIL_RENDERING_HOST}/data-article/render-template`;
const TEMPLATES_LIST_URL = `${EMAIL_RENDERING_HOST}/info/templates/`;

```

apps/newsletters-ui/src/app/components/TemplatePreview.tsx
```diff
@@ -0,0 +1,89 @@
+import RefreshIcon from '@mui/icons-material/Refresh';
```
phillipbarron, line 1: what github says :)

phillipbarron (approved):
This is really great, awesome job.

Tiny nit is that would could adfd a few pixels of padding below above the preview as it's right up against the series tag notification which looks a little busy to my eye.

love it!

dblatcher:
> Tiny nit is that would could adfd a few pixels of padding below above the preview as it's right up against the series tag notification which looks a little busy to my eye.

👍 I'm also thinking we should let this page be wider than the others - the `ContentWrapper` limits the page width to 'lg', but I could add a width prop so this page can be 'xl'.

dblatcher:
> > > Tiny nit is that would could adfd a few pixels of padding below above the preview as it's right up against the series tag notification which looks a little busy to my eye.
> > 
> > 
> > 👍 I'm also thinking we should let this page be wider than the others - the `ContentWrapper` limits the page width to 'lg', but I could add a width prop so this page can be 'xl'.
> 
> Yeah, it is a little squashed - give it a try 👍

Pushed some layout changes - I also used a MUI elevated Card instead of having the outline on the preview container

Screenshot 2023-07-20 at 10 40 14

PR #225, opened 2023-07-21
Update route / actions grouping and Nav styling

phillipbarron:
## What does this change?

Aligns the routes with the menu options and groups actions based on state. 

Adds an update to the Nav styling to make make it clearer which section you are in

## How to test

Make sure everything still works. Local testing looks ok but it would be good to have a  second pair of eyes on this

Yep

## Images

Screenshot 2023-07-24 at 07 43 19
ally only applicable to UI changes, check the boxes if you are satisfied that your changes pass these tests -->

### DIFF
--- a/apps/newsletters-ui/src/app/routes/launched.tsx
+++ b/apps/newsletters-ui/src/app/routes/launched.tsx
@@ -3,7 +3,7 @@ import { NewsletterDetailView } from '../components/views/NewsletterDetailView';
 import { NewsletterEditView } from '../components/views/NewsletterEditView';
 import { NewsletterJsonEditView } from '../components/views/NewsletterJsonEditView';
 import { NewslettersListView } from '../components/views/NewslettersListView';
-import { PreviewView } from "../components/views/PreviewView";
+import { PreviewView } from '../components/views/PreviewView';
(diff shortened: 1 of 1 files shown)


dblatcher:
Afraid's one more `NavigationButton` to update in "apps/newsletters-ui/src/app/components/views/RenderingOptionsView.tsx"

dblatcher:
We may need to add '/launched' and '/launched/*' to the paths handled by `handleUiRequest` in "apps/newsletters-api/src/register-ui-server.ts" - we needed to do this in production so the browsing directly (as opposed to using a navigation action within the app) to a page in the UI will be handled by the router.

phillipbarron:
> RenderingOptionsView

Ahh interesting - thanks - that was not on my radar at all
let me check

phillipbarron:
> We may need to add '/launched' and '/launched/*' to the paths handled by `handleUiRequest` in "apps/newsletters-api/src/register-ui-server.ts" - we needed to do this in production so the browsing directly (as opposed to using a navigation action within the app) to a page in the UI will be handled by the router.

dblatcher (approved):

PR #226, opened 2023-07-25
Dynamic Image Signing

phillipbarron:
## What does this change?

Allows dynamic signing of unsigned images allow us to direct users to the s3 uploader and for us to return a signed image via the readonly api. 

This also adds a config service which will fetch values from SSM. Values have been set for all environments both for the tool and api. The config service adds the values to state to prevent repeated calls to SSM

**EDIT** - the API will return the actual unsigned URLs by default. If the `ENABLE_DYNAMIC_IMAGE_SIGNING` env var is set to true, the API can be asked for the data with the supported images signed using a query parameter:
http://localhost:3000/api/newsletters?signImages=true

## How to test

Checkout the code and run the setup script to add the appropriate environment variables to your .env.local (**remember this will overwrite your config - backup if you have anything non default defined**)

Tested locally & in CODE

```bash
./scripts/setup.sh
```

- To test locally, update the `ENABLE_DYNAMIC_IMAGE_SIGNING` value to true. It is set to false by default as this is its setting in the tool. The readonly api is set to dynamically sign images. 
- Upload an asset in s3 uploader and copy the vanity url
- Run the Newsletter tool and edit the rending options of a launched newsletter
- check the api response (**EDIT**  with the query param added, ie http://localhost:3000/api/newsletters/**{identity-name}**?signImages=true )for the Newsletter after adding the image - you should see that the image has been signed 

Signing will only work for images served from Fastly (uploads.guim.co.uk) - there is logic in place to ignore all other hosts. This might need refining

#### original data
Screenshot 2023-07-25 at 21 29 14

#### api response with "signImages=true"
Screenshot 2023-07-25 at 21 27 59

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ To run the UI and API locally with the default options:
 
 ### Run the app
 
+fetch some `frontend` credentials from [Janus](https://janus.gutools.co.uk/credentials?permissionId=frontend-dev&tzOffset=1) 
+
 `npm run dev`
 
 See the documentation for the [API](apps/newsletters-api/README.md) for the configuration options.
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ To run the UI and API locally with the default options:
 
 ### Run the app
 
-fetch some `frontend` credentials from [Janus](https://janus.gutools.co.uk/credentials?permissionId=frontend-dev&tzOffset=1) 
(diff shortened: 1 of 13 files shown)


dblatcher (approved):
Look good - might be worth updating the description for future reference.

PR #228, opened 2023-07-27
Add banner and illustration card collection to wizard

phillipbarron:
## What does this change?

Adds collection of URLs for the mainBannerURL and illustrationCard to the data collection and rendering options wizard.

The Main Banner has been added in to the Header section along with Date & Standfirt. A new image has been added to highlight the three elements.

For both new options, users are invited to upload images via the s3 uploard (to which we provide a link)

## How to test

Test we can add & remove these elements and that they are inlcuded in the API response. 

There's a usibility issue in that we have reached 15 steps again in the data colletion Wizard. We have an existing trello ticket to group some of the options.We might think about playiong that ticket soon.

## Images

Screenshot 2023-07-27 at 21 15 34

Screenshot 2023-07-27 at 21 16 34


libs/newsletter-workflow/src/lib/steps/newsletterData/illustrationCardLayout.ts
```diff
@@ -0,0 +1,53 @@
...
+import { executeModify } from '../../executeModify';
+import { executeSkip } from '../../executeSkip';
+import { getStringValuesFromRecord } from '../../getValuesFromRecord';
+import { regExPatterns } from '../../regExPatterns';
+import { formSchemas } from './formSchemas';
+
+const markdownTemplate = `
+## Add an illustration card for of {{name}}
+
+Illustration card images are used to promote newsletters on the newsletters page and on thrashers.
```
dblatcher, line 13: We don't actually use the illustrationCard image in thrashers as such - we might use the same asset as the 'mobile' version of the graphic, but on most views it needs to be a square image. it might be best not to mentions thrashers here.

dblatcher (approved):
just one suggestion on the copy

libs/newsletter-workflow/src/lib/steps/newsletterData/illustrationCardLayout.ts
phillipbarron: Ahh cheers - misunderstood that use 👍

PR #230, opened 2023-07-31
footer email and onward path

dblatcher:
## What does this change?

 - Changes the schema for `renderingOptions`, so the `contactEmail` is optional and   for `readMoreSections`:
   - the "url" field is optional
   - there is a new optional "onwardPath" field (with regex validation)
 - Derives a "data collection" version of the `renderingOptions` schema, so that in `readMoreSections` for new drafts: 
   - the "url" field is excluded
   - the "onwardPath" field is required

This will allow us to change the schema used for live data without invalidating any of the existing newsletters, and also collect the 'correct' data for any new newsletters.

After this change is merged, we can:
 - update the existing data on PROD and CODE to replace the  `renderingOptions.readMoreSections[].url` values with `renderingOptions.readMoreSections[].onwardPath` values, using a custom update script in  https://github.com/guardian/newsletters-migration-tool
 - do another PR in this repo to remove `URL` and make `onwardOath` required 

## How to test

functionality - on the rendering options wizard (eg http://localhost:4200/drafts/newsletter-data-rendering/7009), the 'Read More Sections' step will prompt the user to input the 'onwardPath'. With this value populated, a completed draft can still be launched.

data integrity - deploying this branch to CODE should not change the CODE data response - IE no missing newsletters.

The updated data model for `renderingOptions` can be consumer by email-rendering - the onwardPath will be a string in the correct format for the relative URL needed for the readmore section.

Users of the tool will no longer be required to enter a custom contact email for every new newsletter, but are informed that leaving it blank will result in the default contact email ("newsletters@thegaurdian.com") being used in email rendering.

Changing the `newsletter-data-schema` can result in data being excluded from PROD if the existing data does not match the new schema.

We could do with integration tests to mitigate this, but manual testing should be sufficient for now.

### DIFF
--- a/libs/newsletters-data-client/src/lib/draft-newsletter-data-type.ts
+++ b/libs/newsletters-data-client/src/lib/draft-newsletter-data-type.ts
@@ -4,7 +4,9 @@ import { metaDataSchema } from './meta-data-type';
 import {
 	newsletterDataSchema,
 	onlineArticleSchema,
+	readMoreSectionSchema,
 	regionFocusEnumSchema,
+	renderingOptionsSchema,
 } from './newsletter-data-type';
 import { nonEmptyString } from './zod-helpers';
 
@@ -31,6 +33,33 @@ export function isDraftNewsletterDataWithMeta(
 		.extend({ meta: metaDataSchema })
 		.safeParse(subject).success;
 }
+
+/**
+ * A version of the renderingOptionsSchema
+ * for use when defining new drafts.
+ *
+ * In this version ,the readMoreSections require
+ * the `onwardPath` and  exclude the `url`.
+ * both are optional in the 'real 'schema.
+ */
+export const dataCollectionRenderingOptionsSchema =
+	renderingOptionsSchema.merge(
+		z.object({
+			readMoreSections: readMoreSectionSchema
+				.pick({
+					subheading: true,
+					wording: true,
+					onwardPath: true,
+				})
+				.merge(
+					z.object({
+						onwardPath: nonEmptyString(),
+					}),
+				)
+				.array()
+				.optional(),
+		}),
+	);
 
 /**
  * The schema for collecting data for new drafts.
--- a/libs/newsletters-data-client/src/lib/draft-newsletter-data-type.ts
+++ b/libs/newsletters-data-client/src/lib/draft-newsletter-data-type.ts
@@ -8,7 +8,7 @@ import {
 	regionFocusEnumSchema,
 	renderingOptionsSchema,
 } from './newsletter-data-type';
-import { nonEmptyString } from './zod-helpers';
+import { nonEmptyString, urlPathString } from './zod-helpers';
 
 export const draftNewsletterDataSchema = newsletterDataSchema.deepPartial();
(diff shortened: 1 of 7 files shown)


libs/newsletters-data-client/src/lib/newsletter-data-type.ts
```diff
@@ -32,10 +33,19 @@ export const singleThrasherLocation = z
 	.optional();
 export type SingleThrasherLocation = z.infer<typeof singleThrasherLocation>;
 
+export const readMoreSectionSchema = z
+	.object({
+		subheading: nonEmptyString().describe('read more subheading'),
+		wording: nonEmptyString().describe('read more wording'),
+		url: z.string().url().describe('read more url'),
```
alinaboghiu, line 40: did we want this to be .optional() as well?

libs/newsletters-data-client/src/lib/draft-newsletter-data-type.ts
```diff
@@ -32,6 +34,33 @@ export function isDraftNewsletterDataWithMeta(
...
+ * both are optional in the 'real 'schema.
+ */
+export const dataCollectionRenderingOptionsSchema =
+	renderingOptionsSchema.merge(
+		z.object({
+			readMoreSections: readMoreSectionSchema
+				.pick({
+					subheading: true,
+					wording: true,
+					onwardPath: true,
```
alinaboghiu, line 52: is this necessary if we are doing the merge below?

alinaboghiu:
Running this locally to test onwardsPath 
* for an existing draft the field has a star, suggesting it might be required, but it isn't
* for a new draft, your comment suggests it should be required, but I am able to omit it 
* I cannot see the url field at all for existing drafts (maybe this is intentional)
* I am not able to edit onwards journeys for existing launched newsletters as far as I can see, I think this is fine

libs/newsletters-data-client/src/lib/newsletter-data-type.ts
dblatcher: yes - that should be optional!

dblatcher:
> Running this locally to test onwardsPath
> 
> * for an existing draft the field has a star, suggesting it might be required, but it isn't
> * for a new draft, your comment suggests it should be required, but I am able to omit it

It was "required" on the schema, but the regex pattern was satisfied by an empty string and omitting the value in the field sets the value to `''` - i've updated the regex so the path has to be at least `'/'`

> * I cannot see the url field at all for existing drafts (maybe this is intentional)

Yes, that's intentional - we need the url field to be allowed to keep the existing data valid, but are phasing it out. The only code that consumes the value is the WIP 'data driven article' route in email rendering - which will be updated to use 'onwardPath' instead

> * I am not able to edit onwards journeys for existing launched newsletters as far as I can see, I think this is fine

It should show up as a field  in the EditRenderingOptionsForm  (/launched/rendering-options/{identityName}) when you add an item to 'readMoreSections' -  but the change won't affect the url in the preview content yet as we need to update email-rendering if that's what you mean?

alinaboghiu (approved):

alinaboghiu (approved):

PR #231, opened 2023-08-09
Dblatcher add pending status

dblatcher:
## What does this change?

 - adds new value of 'pending' to the `status` enum
 - makes 'pending' the default initial status after newsletters are launched
 - updates `deriveLegacyNewsletter` to set `paused` to false on the legacy newsletters if the `newsletterData.status` is 'pending
 - adds more test coverage

## How to test

locally:
 - the status for the hard-coded test newsletter [dram-security](http://localhost:4200/launched/dram-security) has it's status set to "pending" - the [legacy API page](http://localhost:4200/api/legacy/newsletters) will show the boolean "paused" as true for dram-security
 - create a new draft and launch it - initial status will be 'pending' and the [legacy API page](http://localhost:4200/api/legacy/newsletters) will show the boolean "paused" as true for your new draft
 - the "pending" status can be selected from the status options on the edit newsletter page, eg http://localhost:4200/launched/edit/electric-bicycle

Initial status of any new newsletters should be more informative - we can distinguish 'paused' and 'pending'.

Allows further work relating to recording async tasks being made to the newsletter (eg tag set-up, Braze campaigns) before status can move from 'pending' to 'live.

The UI is not being used by external stakeholder in production and the new data model in only consumed by email-rendering (for which the `status` field is not relevant), so should be no risk.

### DIFF
--- a/apps/newsletters-api/src/services/storage/inMemoryStorageInstance.ts
+++ b/apps/newsletters-api/src/services/storage/inMemoryStorageInstance.ts
@@ -8,7 +8,6 @@ export const makeInMemoryStorageInstance = () =>
 			signUpDescription:
 				'Quod nobis aperiam aperiam ipsum nobis officiis explicabo molestias quibusdam.',
 			frequency: 'Weekly',
-			status: 'paused',
 			restricted: false,
 			emailConfirmation: false,
 			signUpEmbedDescription: "We'll send you Avon Recumbent weekly",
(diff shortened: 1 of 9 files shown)


phillipbarron (approved):
Looks good to me 👍

dblatcher:
I had a second thought about this - maybe the "pending" status should be mapped to `cancelled=true` on the legacy data model. Its not semantically right, but it would represent a state for a newsletter that exists in the API, but should not appear on the site and should not have any subscribers yet (so we don't need the MMA page to allow people to unsubscribe).

phillipbarron:
> I had a second thought about this - maybe the "pending" status should be mapped to `cancelled=true` on the legacy data model. Its not semantically right, but it would represent a state for a newsletter that exists in the API, but should not appear on the site and should not have any subscribers yet (so we don't need the MMA page to allow people to unsubscribe).

Yeah good idea

dblatcher:
> > I had a second thought about this - maybe the "pending" status should be mapped to `cancelled=true` on the legacy data model. Its not semantically right, but it would represent a state for a newsletter that exists in the API, but should not appear on the site and should not have any subscribers yet (so we don't need the MMA page to allow people to unsubscribe).
> 
> Yeah good idea

On second (third?) thoughts, that would mean that the pending newsletters can't be signed up to yet, which might not be the desired behaviour. Maybe one for another PR after some discussion.

PR #232, opened 2023-08-14
Add statuses for async workflows and set them to a default on creation

phillipbarron:
## What does this change?

Adds four new properties to the newsletters data model

```
brazeCampaignCreationsStatus
ophanCampaignCreationsStatus
tagCreationsStatus:
signupPageCreationsStatus
```

These are set to `NOT_REQUESTED` at the point of creating a draft. 

The intent is that the values will be updated as those activities are completed and this will allow us to surface launch progress in the UI. 

In this first PR, the values are set to Optional. We can merge this PR, then backfill the existing draft & launched newsletters before creating a  final PR to set the property to required. 

## How to test

Create a newsletter (locally or when deployed) both using inMemory & s3 storage and then check that the values exist and are as expected.

Values exist and are as expected

Nothing else in the app cares about these values so there is low risk in this change


libs/newsletters-data-client/src/lib/draft-storage/InMemoryDraftStorage.ts
```diff
@@ -33,6 +33,10 @@ export class InMemoryDraftStorage implements DraftStorage {
 			listId: this.getNextId(),
 			creationTimeStamp: Date.now(),
 			meta: undefined,
+			brazeCampaignCreationsStatus: 'NOT_REQUESTED',
+			ophanCampaignCreationsStatus: 'NOT_REQUESTED',
+			tagCreationsStatus: 'NOT_REQUESTED',
+			signupPageCreationsStatus: 'NOT_REQUESTED'
```
dblatcher, line 39: The values don't seem to get populated when running locally 🤔 . Not sure why. But It would probably be better to use the `defaultNewsletterValues` in libs/newsletters-data-client/src/lib/draft-to-newsletter.ts - that way you won't need to change both the DraftStorage implementations.

phillipbarron: Right - I like writing less code but I guess I will need to update the schema to exclude these properties for the draft, otherwise it would always fail validation in the draft state.

I will take a look

phillipbarron:
> looks good - but I think we should set the new values at the point of lauch using the defaults constant so it doesn't need to be handled in the storage classes.

I had imagined that there might be circumstances in which we could either mark or infer the status of these at draft time. Perhaps having common draft creation status might be another way around this. I will have a think about it. We can be fairly sure that the Braze & ophan stuff has not happened but the tags & sign-up 🤷‍♂️

libs/newsletters-data-client/src/lib/draft-storage/InMemoryDraftStorage.ts
phillipbarron: I have moved the default values logic into the workflow `executeCreate` function to reduce duplication. I think it's simpler to set these at draft creation time and this gives us the option of marking / inferring completion of some async workflow tasks at data collection time. 

Regarding the failure to populate, I cannot recreate. Both before and after the change, creating a draft (after setting a  name), results in the values being set. I have tested this in both inMem & s3. Happy to jump on  a call to see if we can troubleshoot together?

dblatcher:
> > looks good - but I think we should set the new values at the point of lauch using the defaults constant so it doesn't need to be handled in the storage classes.
> 
> I had imagined that there might be circumstances in which we could either mark or infer the status of these at draft time. Perhaps having common draft creation status might be another way around this. I will have a think about it. We can be fairly sure that the Braze & ophan stuff has not happened but the tags & sign-up 🤷‍♂️

We could do both - if (for example) the sign up page had been created pre-launch, the status could be set on the draft - the default of "NOT_REQUESTED" would be overriden by the populated value on the draft in "draft-to-newsletter".

libs/newsletters-data-client/src/lib/draft-storage/InMemoryDraftStorage.ts
dblatcher: I was using the existing hard-coded test drafts and running through the launch step to see if the values were set on the launched newsletters - but I see now why that wouldn't work - you are setting the values at draft creation, and the test drafts don't have them set, so they aren't set on the newsletters resulting from the launch.

related point - I noticed I left a bug in the `withDefaultNewsletterValuesAndDerivedFields` in "draft-to-newsletter" - the value for "status" is only being set if the draft has a value for `renderingOptions` - if not the draft fails the 'is draft ready' check

phillipbarron: Ahh right - I will update the hard-coded ones
Good spot

dblatcher (approved):

PR #233, opened 2023-08-15
fix defaulting on drafts

dblatcher:
## What does this change?

Fixes and simplifies the logic in draft-to-newsletter so the `status` field will always be populated and the logic for 'renderingOption' is documented.
extends tests to account for the fixed bug.

## How to test

Changes have test coverage `nx run newsletters-data-client:test`

### DIFF
--- a/libs/newsletters-data-client/src/lib/draft-to-newsletter.ts
+++ b/libs/newsletters-data-client/src/lib/draft-to-newsletter.ts
@@ -32,9 +32,6 @@ export const withDefaultNewsletterValuesAndDerivedFields = (
 ): DraftNewsletterData &
 	Pick<NewsletterData, NewsletterFieldsDerivedFromName> => {
 	const derivedFields = deriveNewsletterFieldsFromName(draft.name ?? '');
-	//prevent an explicit undefined status on the draft overriding the default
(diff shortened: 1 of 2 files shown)


libs/newsletters-data-client/src/lib/draft-to-newsletter.ts
```diff
@@ -28,36 +32,38 @@ export const withDefaultNewsletterValuesAndDerivedFields = (
...
-	if (draft.renderingOptions) {
-		return {
-			...defaultNewsletterValues,
-			...derivedFields,
-			...draft,
-			//prevent an explicit undefined status on the draft overriding the default
-			status: draft.status ? draft.status : defaultNewsletterValues.status,
-			renderingOptions: {
+	//prevent an explicit undefined status on the draft overriding the default
+	const getDefaultedStatus = () =>
```
phillipbarron, line 36: does this need to be a function? I may be missing something but this looks the same as:

```
draft.status ?? defaultNewsletterValues.status
```

We could inline that rather than declaring a function right

dblatcher: 👍  done

phillipbarron (approved):
👍

PR #235, opened 2023-08-17
Dblatcher automation rendering options

dblatcher:
## What does this change?

Moves the schema / type definitions into a subfolder with smaller files

Adds a new optional renderingOptions property for automated fronts sections.

## How to test

Run locally  (without USE_LOCAL_EMAIL_RENDERING  set to true)
`npm run dev`

go to the rendering options editor on the tool, eg "http://localhost:4200/launched/rendering-options/bmx-tesla"

Define some automated sections - the tool will post the data to local email rendering, which will return a sample article with the content from the front - eg:
- subheading: Sport section
- insertPosition: 1
- frontPath: /uk
- sectionName: sport
- maxArticles: 3

Users can define automated sections in the tool.

The `AutomatedFrontSection.builderFunctionName` property represents the name of a "part builder" function in email-rendering. The value is typed as a string in the data model, but there is a hard coded list of the available options for data collection. This works, but might not be ideal data-design. It would be better to display user-friendly descriptions for the functions in the UI.

The user gets no feedback if they define an 'invalid' section - eg  pointing to a front or front collection that doesn't exist - the preview will just not include a section. We could do extra work on the the email-rendering endpoint to return a list of warnings as well as the renderded HTML.

## Images

Screenshot 2023-08-17 at 09 36 55


libs/newsletters-data-client/src/lib/schemas/data-collection-schema.ts
```diff
@@ -0,0 +1,84 @@
...
+	renderingOptionsSchema,
+} from './rendering-options-data-type';
+
+// TO DO - can we fetch the list of supported functions from email-rendering
+// and dynamically create the schema? is that overkill?
+// NOTE - data validation schema types partBuilderFunction as non-empty string
+// seemed too rigid to use the names in the data schema - might need a data migration
+// to add/remove/deprecate options. Defining the options for data collection only
+// give enough flexibility
+const builderFunctionNames = [
```
phillipbarron, line 20: This also makes me think we need to have a solution for display lables for this kind of content. Our forms all look a bit WIP because there is so much camelCase on display. Not one for this PR but an important health ticket.
There is an example somewhere of extending Zod schemas to add a label property that can be used for UI purposes

phillipbarron (commented):
Looks great though I am not able to sucessfully render locally. Not clear why. Have both the branches checked out & using local rendering

dblatcher:
> Looks great though I am not able to sucessfully render locally. Not clear why. Have both the branches checked out & using local rendering

Did you use `PORT=3010` when running email rendering? If it wasn't that, we should take a look together.
(side note - maybe we should change the default port on email rendering)

libs/newsletters-data-client/src/lib/schemas/data-collection-schema.ts
dblatcher: I think it's set up so the Zod field's `description` property is used as the field label (if set - if not it falls back to the field Id). 
But, agreed, we probably could handle that better.

alinaboghiu:
Just a small note to say please try to break down your PRs a little more, this is quite intimidating to review as a single change. In this case moving the schema and refactoring could have perhaps been a separate PR.

alinaboghiu (approved):

PR #237, opened 2023-08-22
Dblatcher email send service

dblatcher:
## What does this change?

Adds new .env variables to for configuring the sending of email notifications:
 - ENABLE_EMAIL_SERVICE: whether the notifications are enabled 
 - TEST_EMAIL_RECIPIENTS: a list of email address to receive any notifications sent if the stage is not PROD (developers can set this to their own email address to test sending messages locally - in PROD, the recipient list is determined by the type of message)

Adds a new lib ("email-builder"), which exports a `sendEmailNotifications` function to act as an internal interface for sending email notifications from the newsletters-api application. The function takes the follow arguments:
 - a `MessageParams` object describing the message to be sent
 - an instance of the AWS `SESClient`
 - an `EmailEnvInfo` object describing the configurations values above

Adds an `SESClient` and `EmailEnvInfo` to the properties of the `DraftService` and `LaunchService`, so that `sendEmailNotifications` can be called within the newsletter-workflow library

Makes calls to `sendEmailNotifications` when  a new draft is created and when a newsletter is launched.

## How to test

locally:
set env vars in .env.local, including your email address in TEST_EMAIL_RECIPIENTS:
```
# toggle whether to send email notifications
ENABLE_EMAIL_SERVICE=true

# The email addresses to which all emails are sent to when not on PROD.
# use semi-colons to separate the values.
TEST_EMAIL_RECIPIENTS=firstname.lastname@guardian.co.uk;
```

Run the tool as normal, `npm run dev` :
 - hitting the test-email endpoint, http://localhost:3000/email/bmx-tesla will trigger sending the "test" email to your test recipients
 - creating a new draft from  in [the draft wizard](http://localhost:4200/drafts/newsletter-data) (IE completing the first step where you specify the name) will trigger a "new draft" email
 - launching an draft from [the launch wizard](http://localhost:4200/drafts/launch-newsletter/5670) will trigger a "newsletter launched" email

Provides a convenient, flexible  and robust interface for sending notifications in response to user interactions.

The content of the message template in the "email-builder" lib are for testing only - the actual messages and when they need to be sent are still TBC.

Should be safe - as long as the ENABLE_EMAIL_SERVICE is not set to true in the CDK code, the `sendEmailNotifications` calls on PROD (and CODE) will exit early with no action.

We don't yet know how the recipients of real messages on PROD would be set - eg they could be hard coded for each message type, but we might want to make this configurable.


apps/newsletters-api/env.local.example.txt
```diff
@@ -39,3 +39,10 @@ USER_PERMISSIONS={"software.developer@guardian.co.uk": 0}
 # whether to fetch from a local instance of email rendering on port 3010
 # instead of the PROD service.
 # USE_LOCAL_EMAIL_RENDERING=true
+
+# toggle whether to send email notifications
+ENABLE_EMAIL_SERVICE=false
+
+# The email addresses to which all emails are sent to when not on PROD.
+# use semi-colons to separate the values.
+# TEST_EMAIL_RECIPIENTS=firstname.lastname@guardian.co.uk;other.person@guardian.co.uk;
```
phillipbarron, line 48: Do we think this allows sufficient control over who receives emails?

I have been away so perhaps this has already been discussed? In this model, everyone gets every message - I can see that _we_ might want every message, but CP might not? 

We might want to think about whether we define email groups in config that will allow us to include / exclude recipients programatically - That could come later, though - not necessarily for this PR

phillipbarron (approved):
Looks great - left a comment regarding email groups in config, but we can think about that later if required

apps/newsletters-api/env.local.example.txt
dblatcher: The TEST_EMAIL_RECIPIENTS are only used when not on PROD (IE in DEV or CODE). In PROD, the recipients are determined by what type of messages it is - the function for building the message passes the list of `prodRecipients` to `getMessageConfig`.
At them moment, all the messages would send to 'newsletters.dev@guardian.co.uk' on PROD if the service was enabled, but we'll be able set appropriate recipients for different messages.

phillipbarron: Ahh gotcha - misread the intent there 👍

PR #238, opened 2023-08-24
Dblatcher headline section rendering options

dblatcher:
## What does this change?

Updates the data model with the x2 optional booleans in the renderingOptions:
 - darkHeadlineBackground
 - displayNewsletterName

## How to test

New options will be included on the Edit rendering options page 

Will be able to use the new options

is safe


phillipbarron (approved):
lgtm

PR #239, opened 2023-08-31
Tiered Edit Access

phillipbarron:
## What does this change?

This change introduces 4 new edit access levels:

- TagEditor
- BrazeEditor
- OphanEditor
- SignUpPageEditor

The edit form presented to a user in based on their access level. `TagEditors` may edit tags and the tag workflow status, `BrazeEditors` may edit the Braze values and workflow status etc.

This allows us to limit access to the edit view for a newsletter based on user context and should facilitate our async update workflows which request an update from a team to update the status of some activity outside the application. 

## How to test

- Set the access level to test in the .env.local and run the application (this file is evaluated at start time so you will need to restart for each test)
- Verify that the edit form shows the limited view (based on your selected access level)

That views and updates are limited based on user permission status

Should be OK - tested locally

### DIFF
--- a/apps/newsletters-api/src/app/authorisation.ts
+++ b/apps/newsletters-api/src/app/authorisation.ts
@@ -1,0 +1,43 @@
+import type { FastifyRequest } from 'fastify/types/request';
+import type { UserProfile } from '@newsletters-nx/newsletters-data-client';
+import {
+	getUserEditSchema,
+	isPartialNewsletterData,
+} from '@newsletters-nx/newsletters-data-client';
+import { permissionService } from '../services/permissions';
+
+export const hasEditAccess = async (
+	profile: UserProfile | undefined,
+): Promise<boolean> => {
+	if (!profile) return false;
+	const permissions = await permissionService.get(profile);
+	const { editTags, editNewsletters, editSignUpPage, editBraze, editOphan } =
+		permissions;
+	return [editTags, editNewsletters, editSignUpPage, editBraze, editOphan].some(
+		(permission) => permission,
+	);
+};
+
+export const isAuthorisedToUpdateNewsletter = async (
+	profile: UserProfile | undefined,
+	request: FastifyRequest,
+): Promise<boolean> => {
+	// get the users permissions and check if they have editNewsletters permission - if so, return true (because they can do whatever they want)
+	// if not, check what edit permission they have and then check the keys in the payload to see if they are allowed to update them
+	// if they are allowed to update them, return true, otherwise return false
+	if (!profile) return false;
+	const permissions = await permissionService.get(profile);
+	const { editNewsletters } = permissions;
+	if (editNewsletters) return true;
+
+	const { body: modifications } = request;
+
+	if (!isPartialNewsletterData(modifications))  {
+		throw new Error('Invalid newsletter data');
+	}
+	const updateKeys = Object.keys(modifications);
+
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-argument,@typescript-eslint/no-unsafe-call
+	const userEditSchema = Object.keys(getUserEditSchema(permissions));
+	return updateKeys.every((key) => userEditSchema.includes(key));
+};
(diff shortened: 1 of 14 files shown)


apps/newsletters-api/src/app/authorisation.ts
```diff
@@ -0,0 +1,42 @@
...
+	const { editTags, editNewsletters, editSignUpPage, editBraze, editOphan } =
+		permissions;
+	return [editTags, editNewsletters, editSignUpPage, editBraze, editOphan].some(
+		(permission) => permission,
+	);
+};
+
+export const isAuthorisedToUpdateNewsletter = async (
+	profile: UserProfile | undefined,
+	request: FastifyRequest,
```
dblatcher, line 23: I would tend to extract the body (and maybe parse it using isPartialNewsletter) from the request and use that as the argument rather than sending the whole request to this function.

That way we only need to think about the API using Fastify within the modules that deal with the routes - it would also make the function easier to test

apps/newsletters-api/src/app/authorisation.ts
```diff
@@ -0,0 +1,42 @@
...
+	if (!profile) return false;
+	const permissions = await permissionService.get(profile);
+	const { editTags, editNewsletters, editSignUpPage, editBraze, editOphan } =
+		permissions;
+	return [editTags, editNewsletters, editSignUpPage, editBraze, editOphan].some(
+		(permission) => permission,
+	);
+};
+
+export const isAuthorisedToUpdateNewsletter = async (
```
dblatcher, line 21: Just a suggestion, but maybe this function's name should make it clear that it is checking if the user is authorised to make this partical set of updates rather than able to update generally - maybe `isAuthorisedToMakeRequestedNewsletterUpdate` ?

apps/newsletters-api/src/app/routes/newsletters.ts
```diff
@@ -79,99 +84,121 @@ export function registerReadNewsletterRoutes(app: FastifyInstance) {
...
 
-		const { newsletterId } = req.params;
-		const { body: modifications } = req;
-		const newsletterIdAsNumber = Number(newsletterId);
+	app.patch<{
+		Params: { newsletterId: string };
+		Body: unknown;
+	}>(
+		'/api/newsletters/:newsletterId',
+		{ preValidation: hasAccessHook },
```
dblatcher, line 107: I've not seen this feature before! Not necessarily one for this PR, but it looks like we could use another hook to do the 'Non numeric id provided' check too

libs/newsletters-data-client/src/lib/zod-helpers/user-data-schema.ts
```diff
@@ -0,0 +1,60 @@
...
+		});
+	}
+	if (editSignUpPage) {
+		return newsletterDataSchema.pick({
+			signupPageCreationsStatus: true,
+			signupPage: true,
+			signUpDescription: true,
+		});
+	}
+	throw new Error('No permissions found');
```
dblatcher, line 59: Do we want this to throw an Error when the user doesn't have any permissions? Might that cause a run time error when it is used in a component? 

Would it work to return an empty schema as a fallback - IE  `newsletterDataSchema.pick({})` - that way the function should be safe and would reflect that there are no fields the user is allowed to edit

dblatcher (commented):
looks good - made a couple of comments, but nothing major

apps/newsletters-api/src/app/routes/newsletters.ts
phillipbarron: Yeah - found it when looking at middleware for Fastify - basically it uses hooks instead and there are a bunch of events you can hook in to.

apps/newsletters-api/src/app/authorisation.ts
phillipbarron: Good shout

libs/newsletters-data-client/src/lib/zod-helpers/user-data-schema.ts
phillipbarron: Yep - much better

dblatcher (approved):

PR #244, opened 2023-09-06
Send emails on launch

dblatcher:
## What does this change?

The launch workflow triggers multiple emails, and updates the creation status properties based on whether the email sends were successful or not.

The "Tag creation" email will only be sent if the draft has both required tags defined.

Should have no practical effect in production, as the email service if disabled in the config.

This PR doesn't address the question of how to define / configure the recipients of each message type when sending emails in PROD - still defineid in code in the message modules.

## How to test

Locally:
enable email service and set yourself to the test recipient:
```
# toggle whether to send email notifications
ENABLE_EMAIL_SERVICE=true

# The email addresses to which all emails are sent to when STAGE=DEV
TEST_EMAIL_RECIPIENTS=firstname.lastname@guardian.co.uk;
```
Run locally and launch a draft. If the draft has the series tag and composer tag defined, you will get 4x email, otherwise, you will get 3x emails.

After finalising the message contents and recipients, we should be able to turn on the service.

We should think about how this will work in practise, but should be safe.

## Images

Screenshot 2023-09-06 at 17 04 35
Screenshot 2023-09-06 at 17 04 18
Screenshot 2023-09-06 at 17 03 59
Screenshot 2023-09-06 at 17 03 42


libs/email-builder/src/lib/messages/request-tag-creation-message.ts
```diff
@@ -0,0 +1,29 @@
...
+	MessageConfig,
+	MessageContent,
+} from '../types';
+
+export function buildTagCreationRequestMessage(
+	params: MessageAboutNewsletterParams,
+	emailEnvInfo: EmailEnvInfo,
+): { content: MessageContent; messageConfig: MessageConfig } {
+	const { newsletter } = params;
+	const messageConfig = getMessageConfig(
```
phillipbarron, line 15: If we wanted to move email addresses out to configuration, we could use the [config-service](https://github.com/guardian/newsletters-nx/blob/5f4e28320c726b8725404de07963310661f5da82/apps/newsletters-api/src/services/configuration/config-service.ts#L14) - message config could make this call if it were parameterised with the `NewsletterMessageId`

libs/email-builder/src/lib/components/RequestBrazeSetUpMessage.tsx
```diff
@@ -0,0 +1,61 @@
...
+		<MessageFormat
+			title={
+				<>
+					Please create a new Braze campaign for newsletter "
+					{newsletter.identityName}"
+				</>
+			}
+		>
+			<p>
+				The Braze for this newsletter are below - if these need to change for
```
phillipbarron, line 23:
```suggestion
				The Braze values for this newsletter are below - if these need to change for
```

libs/email-builder/src/lib/components/RequestBrazeSetUpMessage.tsx
```diff
@@ -0,0 +1,61 @@
...
+				any reason, please contact the newsletters team.
```
phillipbarron, line 24:
```suggestion
				any reason, please <a href={pageLink}>update in the newsletters tool</a> .
```

We have exposed these elements for `TagEditors` in the form

libs/newsletter-workflow/src/lib/executeLaunch.ts
```diff
@@ -43,6 +36,50 @@ const getExtraValuesFromFormData = (
...
+
+const doNotSendEmail = () =>
+	Promise.resolve({ success: false, error: undefined });
+
+const sendOutEmailsAndUpdateStatus = async (
+	launchService: LaunchService,
+	newsletter: NewsletterData,
+) => {
+	const shouldSendTagCreationRequest =
+		!!newsletter.seriesTag && !!newsletter.composerTag;
```
phillipbarron, line 52: It was not clear to me that a composerTag was a requirement. I can see we need the series tag, but my understanding was that the composer / campaign tags were used for promotion?

phillipbarron (commented):
Looks great - left a couple of comments around email content content and some clarification around tag requirements. We might have a gap between what we are communicating the tag UI (Wizard bit) and the notificaiton conditions.
Looking at this also makes me think we might want a manual send notification UI too.

libs/email-builder/src/lib/messages/request-tag-creation-message.ts
dblatcher: That would work fine in the code, but I'm not sure the parameter store is ideal for storing and managing email addresses - if we need a separate value for the comma-separated list fo email addresses for every message type it add an extra maintenance job when setting up a new message type or updating email address (eg - let's say central production change their contact email and we need to go into AWS and change the param for every messages type that goes to them)

S3 would probably be a better fit - we could just have a single JSON file with the config - something like:
```JSON
{
"REQUEST_TAG_SETUP": {
  "prodRecipients":["central-prod@example.com"],
  "prodCC":["newsletters-editorail@example.com"]
},
"REQUEST_BRAZE_SETUP": {
  "prodRecipients":["production-team@example.com"],
  "prodCC":["newsletters-editorail@example.com"]
}
}
```

libs/newsletter-workflow/src/lib/executeLaunch.ts
dblatcher: That makes sense - the series tag is the only one needed for the newsletter to be functional. 

We could drop the requirement for composerTag to be defined but also update the message template so it only include the 'please create the campaign tag' message if the newsletter has one.

libs/email-builder/src/lib/messages/request-tag-creation-message.ts
phillipbarron: I think the paramter store would work just the same right? there's nothing to stop us having that structure in there - with current implementation, it would need prod and code config options but we can allow an override path in there. 
Anyway, not for this PR

phillipbarron (approved):

PR #246, opened 2023-09-07
correct INTL to INT, add EUR

dblatcher:
## What does this change?

Add's "EUR" to the regionFocus Enum, changes "INTL" to the correct EditionId of "INT"

https://trello.com/c/zPlzeEPJ/632-add-europe-as-regional-focus-to-data

## How to test

All works, new editionId added to the regionFocus UI

can set newsletters to EUR

There are currently no newsletters in PROD that have a regionFocus of "INTL" - should not break
need to update the migration tool too

## Images

Screenshot 2023-09-07 at 16 57 37


phillipbarron (approved):
LGTM

PR #267, opened 2023-09-20
load user permissions data from config-service 

dblatcher:
## What does this change?

Adds an option to the config-service to request a value that was fetch from the store within a give time frame - IE that the value is not too stale.

Sets the request for the email recipient data to specify they be not more than 15 minutes old - IE the service will check for updates if it fetched the value it storer in `state` more than 15 minutes ago.

The user permissions are loaded from the param store at run time (also max age 15 minutes) in the same way as the email permissions data, instead us using the env vars - unless the new "USE_LOCAL_USER_PERMISSIONS" env var is set to 'true'

## How to test

locally 
 - set "USE_LOCAL_USER_PERMISSIONS" to "false"
 - changes in /DEV/newsletters/newsletters-tool/userPermissions will be reflected when you run the app locally
 - changes made to the param store while the app is running will take effect after the specified time (you could change the constant `TIME_BETWEEN_PERMISSIONS_PARAM_CHECKS` constant to 1 minute for the purpose of testing. 

Able to change user permissions without redeploying by changing the param store value

 - Should test on code first.
 - The permissions for CODE and PROD are already in the store - were  being fetched to set the env vars in the stack
 - is 15 minutes too long to wait for a config update? We won't be changing these values often, but the cost of making the request to param store more often shouldn't be significant.
 - if the works, should update the cdk to not add the permissions to the env vars for CODE and PROD - won't be getting used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ See the documentation for the [API](apps/newsletters-api/README.md) for the conf
 
 **NOTE** Any changes to the legacy API data structure should be communicated to the [Data Design](mailto:data.design@theguardian.com) **before** merging to main.
 
-## UI Tool Deployment and Access
+## UI Tool Deployment
 
 The newsletters-tool is deployed to PROD
 https://newsletters-tool.gutools.co.uk/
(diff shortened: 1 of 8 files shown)


phillipbarron (approved):
Love this - really big improvement. Tested locally and all working s expected

PR #274, opened 2023-09-25
Combine tag and sign-up creation emails to cp

phillipbarron:
## What does this change?

Combines the existing Tag creation and sign-up page email requests in to one. Adds a news email config for the CP recipients

## How to test

Update dev / code config to ensure you emails for the `centralProductionRecipients` group. Create and launch a newsletter and ensure that.

Email received as expected on launch

Tested locally so should be OK. Configs all updated to include new group.

@dblatcher - is it expected that we might create a newsletter without Series tags at all (Fronts?) if so, the logic in the email will need a tweak, as will the status updater.

## Images

Screenshot 2023-09-25 at 14 57 24


libs/email-builder/src/lib/components/RenderTagAndSignUpPageCreationMessage.tsx
```diff
@@ -54,24 +65,38 @@ export const RequestTagCreationMessage = ({ pageLink, newsletter }: Props) => {
...
+				newsletter={newsletter}
+				properties={['name', 'signUpHeadline', 'signUpDescription']}
+			/>
+			{embargoDate && (
+				<p>
+					Please add an embargo so that the sign up page should not go live
+					until {embargoDate}.
+				</p>
+			)}
+			{!embargoDate && <p>The page can go live immediately.</p>}
```
dblatcher, line 79: This is more of a product question, but would w want the page to be launched before the newsletter is set to live? That woudl mean the sign-up page would have no form rendered

dblatcher (approved):

libs/email-builder/src/lib/components/RenderTagAndSignUpPageCreationMessage.tsx
phillipbarron: Will clarify with Sophie - thank you

PR #288, opened 2023-10-12
Dblatcher extra field validation

dblatcher:
## What does this change?

Adds some regex validation user-friendly error messaging for the fields:
- signupPage (needs to be the relative path to the article - ie "/" + article.id
- seriesTag (should be the tag path eg "world/series/my-thing")

These were free text fields before. The wizard explained what he values should be, but there is less help on the edit form. 

Users could enter the full url to the sign-up page, the tag name (rather than path) etc and the UI didn't complain

bonus change - the unit test on the schema will output which fixtures failed validation.

## How to test

Try editing a these fields on a lauched newsletter or in the draft wizard. invalid(according the schema) values will be flagged.

Users have more help knowing what values are expected.

The unit test on the schema confirmed that this would not exclude and live newsletters - after catching that "the-long-read" had a trailing space on the end of its series tag (fixed on PROD)


phillipbarron (approved):
Looks good to me

PR #298, opened 2023-11-22
generate composer campaign tag id for notifications

dblatcher:
## What does this change?

Creates a generator for deriving the correct tag ID for a newsletter's "Composer Campaign Tag", in the fomat required by Frontend.

Uses the derived value (with the description for the tag, which is still user specified) in the email to CP.

Adjust the field description for the "Composer Campaign Tag" so the forms prompt the user to provide a Tag Description, not a Tag Id.

## How to test

Set env variable for email service to true, put your email in the dev recipients in param store,
run locally,  launch some newsletters

CP will know the required format for the campaign Tags

Should be safe

## Images
Screenshot 2023-11-22 at 12 06 00


phillipbarron (approved):
LGTM

PR #300, opened 2024-01-04
access hook allows for null values in data

dblatcher:
## What does this change?

Fixes a bug in the user access hook - the UI converts values of `undefined` in the data to `null` (since there is no `undefined` in JSON) before sending the data to the API to allow defined values to be 'cleared'. The API methods account for this and convert the `null`s back to undefined after parsing the JSON.

This wasn't happing in the hook function to check if the user had access, so the API was returning a 403 error when the user tried to set a value to undefined.

Unrelated change - adds a key to the mapped array in the `GeneratedCodeDataPoint` component to resolve a console warning on the details page.

## How to test

on main - using the 'edit newsletter' form to change a 'regional focus' from a defined value to 'undefined' will fail. (error is a 403, but should be a 400)
on this branch - the above will work (assuming user has right permissions).

Users don't encounter the bug

Is safe - the null to undefined function is already type-safe.

### DIFF
--- a/apps/newsletters-api/src/app/routes/newsletters.ts
+++ b/apps/newsletters-api/src/app/routes/newsletters.ts
@@ -90,10 +90,11 @@ export function registerReadWriteNewsletterRoutes(app: FastifyInstance) {
 	) => {
 		const user = getUserProfile(request);
 		const isAuthorised = await hasEditAccess(user.profile);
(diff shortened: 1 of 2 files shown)


phillipbarron (approved):
LGTM 👍

PR #302, opened 2024-01-17
Expose embed description for editors and CP

phillipbarron:
## What does this change?

Updates the edit form to expose the embed description to allow self service updates for editors and CP

## How to test

Run locally and verify that the `signUpEmbedDescription` field is exposed in the edit form and that it can be edited. 

Screenshot 2024-01-17 at 10 24 06

The field is present in the form and editable

CP and users with editorial access (Newsletters editors at present) will be able to update the embed descriptions and these changes will be seen in embeds (after up to 2 hours because of caching). There is no hard limit on the size of the embed description in the tool. It would be possible to add something absurdly long which would make for an unhappy embed. Access is very limited but we might consider introdicing a sensible validated upper limit in future


Georges-GNM (approved):
Checked it locally and confirmed it was now appearing and editable. 

Introducing a valid upper limit would make sense, but I'd agree can wait for the moment

PR #304, opened 2024-01-17
Add digital-cms to codeowners

Georges-GNM:
## What does this change?

Adds digital cms to codeowners as newsletters tooling is part of the team's remit.


phillipbarron (approved):
lgtm

PR #306, opened 2024-02-01
fix deps and update message

phillipbarron:
## What does this change?

Updates the Braze DDR Update email with a reminder to check the plaintext version of the message

## How to test

Test email generation in code 

Low risk

## Images

Screenshot 2024-02-01 at 10 56 21


.vscode/settings.json
phillipbarron, line 1: Should this be in SC?

PR #308, opened 2024-03-01
user profile for emails

dblatcher:
## What does this change?

Adds an optional `user` argument to the `sendEmailNotifications` service and uses it to:
 - identify the user who launched the newsletter (or created the Draft) in the emails to the newsletters team and CP
 - Copy the user on the email to CP

## How to test

Deploy to CODE (and check email recipient config for CODE), create a new Draft, then Launch it. You should be mentioned / copied as described above. (

Email notifications more informative, less questions  asked.

Tell user in advance.


phillipbarron (approved):
Love this,

tested in code - all working beautifully

PR #309, opened 2024-03-05
Reduce instance size from T4g small to micro

cemms1:
## What does this change?

Reduces the instance size of  `newsletters-api` and `newsletters-tool` from T4g small to T4g micro.

## Why?

The apps are hugely over-provisioned (max CPU usage is around 5%).
It would be worth reducing their instance sizes accordingly


davidfurey (approved):
Looks reasonable to me. Is this in preparation for AWS reservations?

PR #319, opened 2024-05-24
More campaign tag related info in UI

dblatcher:
## What does this change?

On the newsletter details page, adds:
- a "generated data point" for the composer campaign tag id
- a tool tip explaining what the 'composer tag(s)' are

On the edit newsletter page, adds some long-form comments:
 -  explaining what the 'composer tag(s)' are
 -  that the composer campaign tag  needs to have an id in the right format to work

There was an update to the display name to the "generated data point" for the composerCampaignTagId - this would also be used in the emails to CP.

## How to test

Just a visual change.

## Images

Screenshot 2024-05-24 at 12 29 10

Screenshot 2024-05-24 at 12 27 59


phillipbarron (approved):
🎉

PR #320, opened 2024-06-10
Remove ophan status as nothing async happens

phillipbarron:
## What does this change?

This removes the async status for Ophan as there is no async process for Ophan so this is just noise. 
The property is left in the model as an optional so we do not get validation errors on the existing newsletters that have this property. We can, at a later date, run a backfill to remove this property from all the existing live newsletters and then remove the property altogether

This also  removed the Ophan element of the launch Wizard. Editorial launch newsletters and don't cre about this. We can edit the Campain code later if required. 

## How to test

Make sure we can create and launch newsletters and that we can view existing newsletters. (Draft and Live), and that the API response is as expected

It still working

Should be fine - nothing externally cares about these properties

## Images

Screenshot 2024-06-10 at 11 30 10


libs/newsletters-data-client/src/lib/schemas/newsletter-data-type.ts
```diff
@@ -163,9 +163,11 @@ export const newsletterDataSchema = z.object({
 	brazeCampaignCreationStatus: workflowStatusEnumSchema.describe(
 		'Braze campaign creation status',
 	),
-	ophanCampaignCreationStatus: workflowStatusEnumSchema.describe(
-		'Ophan campaign creation status',
-	),
+	ophanCampaignCreationStatus: workflowStatusEnumSchema
+		.optional()
+		.describe(
+			'Ophan campaign creation status',
```
dblatcher, line 169: maybe add a comment or update the field description to mark the property as deprecated?

dblatcher (approved):
tested locally - all works!

Suggested adding a comment on newsletter data type to say why this property is optional (just until we get round to removing it properly after a backfill)

libs/newsletters-data-client/src/lib/schemas/newsletter-data-type.ts
phillipbarron: Yes, great idea. Thanks

PR #323, opened 2024-07-03
Remove unsupported Reader Revenue Content bloacks & patch vulnerabilities

phillipbarron:
## What does this change?

The LifeStyle content block is no longer used in Braze but is present in the generated tamplate - this causes an error as we are referencing a non existant block.

This PR removes that reference

There is also a bump to a couple of vulnerabilities

## How to test

Check generated templates no longer have a lifestyle ref

Templates still work

Low risk - this is a manual activity - it's currently broken - this should fix.

### DIFF
--- a/apps/newsletters-ui/src/app/components/GeneratedCodeDataPoint.tsx
+++ b/apps/newsletters-ui/src/app/components/GeneratedCodeDataPoint.tsx
@@ -36,14 +36,7 @@ interface Props {
 	language: string;
 }
 
-export type MerchandisingOverride =
-	| 'default'
-	| 'AUS'
-	| 'Culture'
-	| 'Features'
-	| 'Global'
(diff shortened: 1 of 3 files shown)


dblatcher (approved):
looks good - not one for this PR, but i'm wondering if it would make sense to add some messaging on this feature to the effect that users should speak to BRR if it doesn't work for their newsletter - the output is only 'correct' because it matches how the braze set-up was the last time BRR asked for the a change

phillipbarron:
> looks good - not one for this PR, but i'm wondering if it would make sense to add some messaging on this feature to the effect that users should speak to BRR if it doesn't work for their newsletter - the output is only 'correct' because it matches how the braze set-up was the last time BRR asked for the a change

Yeah it's a good idea - I will add some in - cheers @dblatcher

PR #327, opened 2024-10-18
add cache control headers

dblatcher:
## What does this change?

Adds cache-control (60 second max age) ~~and surrogate control~~ headers to responses from the newsletters api. The headers are currently unset, and since the api is behind fastly, the default TTL of 1 hour is being used.

Note that there are two differently-configured deployments of this repo:
- the "newsletters api" which serves a public, read-only, cached version of the rest API for newsletters data; and 
-  the "newsletters tool" which serves the UI and a read-write version of rest API , both using  google auth to restrict access

## How to test

Deploy this branch to CODE

There should still  be no delay on updates appearing on the "newsletters tool" -  https://newsletters-tool.code.dev-gutools.co.uk/ - as soon as a newsletter is updated, the UI must reflect this.

open https://newsletters.code.dev-guardianapis.com/api/newsletters  ("newsletters api") and check the cache-control header is set

run frontend locally (local frontend uses the CODE newsletters API) - this branch has been modified to reduce the newsletterSignUpAgin refresh time and log out the names of the parsed newsletters: https://github.com/guardian/frontend/tree/dblatcher/newsletter-refresh-debug

see frontend's lit of live newsletters at : http://localhost:9000/email-newsletters.json

make a change to a code newsletter in the tool - eg change the name on https://newsletters-tool.code.dev-gutools.co.uk/launched/edit/alpha

watch local frontend's logs for newsletter refreshes - observe  - the change should come through on a newsletter refresh within 1 minute of your change.

When the Frontend project refreshes its internal cache (currently every 15 minutes), it should receive an update that is not more that five minutes old - ie updates made in the newsletters tool (which update the s3 files which are the data source for the API) should reliably be visible on frontend within 20 minutes.

### DIFF
--- a/apps/newsletters-api/src/app/headers.ts
+++ b/apps/newsletters-api/src/app/headers.ts
@@ -1,0 +1,23 @@
+import type { FastifyReply, FastifyRequest } from 'fastify';
+
+export const getCacheControl = (req: FastifyRequest) => {
+	if (req.routerPath.startsWith('/api/newsletters')) {
+		return {
+			age: '60',
+		};
+	}
+
+	return undefined;
+};
+
+export const setHeaderHook = async (
+	req: FastifyRequest,
+	reply: FastifyReply,
+) => {
+	const cacheControl = getCacheControl(req);
+	if (!cacheControl) {
+		return;
+	}
+
+	void reply.header(`Cache-Control`, `max-age=${cacheControl.age}`);
+};
--- a/apps/newsletters-api/src/app/headers.ts
+++ b/apps/newsletters-api/src/app/headers.ts
@@ -1,4 +1,5 @@
 import type { FastifyReply, FastifyRequest } from 'fastify';
+import { isServingUI } from '../apiDeploymentSettings';
 
 type TtlSettings = {
 	cacheMaxAge: number;
@@ -13,6 +14,11 @@ const newsletterTtl: TtlSettings = {
 export const getCacheControl = (
 	req: FastifyRequest,
 ): TtlSettings | undefined => {
+	// the API instance serving the UI must always provide fresh data
+	if (isServingUI()) {
+		return undefined;
+	}
+
 	if (req.routerPath.startsWith('/api/newsletters')) {
 		return newsletterTtl;
 	}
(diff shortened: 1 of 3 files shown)


apps/newsletters-api/src/app/headers.ts
```diff
@@ -0,0 +1,49 @@
...
+export const setHeaderHook = async (
+	req: FastifyRequest,
+	reply: FastifyReply,
+) => {
+	const cacheControl = getCacheControl(req);
+	if (!cacheControl) {
+		return;
+	}
+
+	// Fastly gives priority to the Surrogate-Control header to determine how often to
```
davidfurey, line 41: Is there any need for the complexity of the split Fastly/Browser cache control headers in this case?

This is often useful when you are making use of Fastly's cache purge APIs - so you can set a much longer surrogate TTL than cache-control TTL, safe in the knowledge that the cache purge API will honour your request to purge regardless of the headers.

But I don't think this service makes use of Fastly's cache purge APIs, so just setting cache-control headers should be sufficient.

apps/newsletters-api/src/app/headers.ts
```diff
@@ -0,0 +1,49 @@
...
+import { isServingUI } from '../apiDeploymentSettings';
+
+type TtlSettings = {
+	cacheMaxAge: number;
+	surrogateMaxAge: number;
+};
+
+const newsletterTtl: TtlSettings = {
+	cacheMaxAge: 60,
+	surrogateMaxAge: 360,
```
davidfurey, line 11: In the PR description you imply this is 5 minutes, but 360s is 6 minutes. Is there any need for the cache time to be this long, would 5s not be sufficient?

apps/newsletters-api/src/app/headers.ts
dblatcher: Probably not - my initial reading of https://docs.fastly.com/en/guides/caching-best-practices#understand-how-cache-control-headers-work was to using surrogate control was recommended generally, but I think I got that wrong.

apps/newsletters-api/src/app/headers.ts
dblatcher: I'm not sure what the ideal time would be but 5 minutes probably is too long. I remember that there was some concern that the newsletters api is effectively public. I'd be a bit wary of going from the default TTL of 1 hour to 5s, but I don't know if I'm being over cautious on that.

Looking at it from the other direction - I don't think there are any uses of the newsletters API that would reasonably require the response to be less than 1 minute old, so maybe 60s is a good cache time?

davidfurey (approved):

PR #330, opened 2024-12-09
add new illustrationSquare property

dblatcher:
## What does this change?

Adds a new optional string property, `illustrationSquare` , which like `illustrationCard` represents the URL to an image that can be rendered by platforms.

Include the new property on the newsletter edit form (but not the new newsletter creation wizard)

## How to test

run the UI and API locally as per the readme.
Propery should be on the edit form (note - it will have the same not-ideal behaviour as the "illustrationCard" field - you cannot type into the field as it will only accept an empty string or valid URL, which must be pasted in - will fix on subsequent PR)

See https://github.com/guardian/newsletters-nx/tree/main/libs/newsletters-data-client#how-not-to-break-the-api for testing that this change will not cause any existing newsletters to fail parsing and be excluded from the API

The property will be available on the API for future work - such as dynamically rendering sign-up block on fronts.

Should be safe as we are only adding an optional property with no current usages.

### DIFF
--- a/apps/newsletters-ui/src/app/components/EditNewsletterForm.tsx
+++ b/apps/newsletters-ui/src/app/components/EditNewsletterForm.tsx
@@ -78,7 +78,7 @@ export const EditNewsletterForm = ({ originalItem }: Props) => {
 					},
 				}}
 				explanations={{
-					illustrationCard: (
+					illustrationSquare: (
 						<Alert severity="info" sx={{ marginBottom: 1, maxWidth: 600 }}>
 							<Typography>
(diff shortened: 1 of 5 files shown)


KaliedaRik (approved):
Testing in local, I can see that this change is effective. Also, all tests pass. Happy to approve!

**Current edit form:**
Screenshot 2024-12-09 at 15 11 52

**Updated edit form:**
Screenshot 2024-12-09 at 15 13 08

**Test results:**
Screenshot 2024-12-09 at 15 34 23

PR #328, opened 2024-10-28
store layouts in s3

dblatcher:
## What does this change?

 - makes a new data type for Editions Layouts - representing the arrangement of newsletters on https://www.theguardian.com/email-newsletters for the different editions (UK, US, AU, INT, EUR) of the page.
 - defines 
   - an abstract class for a service to manage the storage of the editions layouts 
   - an 'in memory' implementation using a static set of initial test data (for local development)
   - an s3 implementation (for CODE / PROD)
 - adds an API route to manage request from the UI to get and update the  Editions Layouts from the UI
 - adds a new permission to control a user's ability to do updates (only available to "developer" profiles for now).
 - builds UI pages to display the layouts and allow user with the permission to edit them (using the existing JsonEditor component - with a view to developing more user-friendly UI later)

### locally
Run locally, with the "developer" permissions set in .env.local (see main readme for setup):
```USER_PERMISSIONS={"software.developer@guardian.co.uk": 0}```

go to http://localhost:4200/layouts to see the main UI - you should be able to create and edit layouts

shut down the app, change the user permission level to :
```USER_PERMISSIONS={"software.developer@guardian.co.uk": 3}``` 

run the app again and you go to go to http://localhost:4200/layouts - you should have read-only access

### on CODE
If the branch is deployed to CODE, the draft layouts created in the [CODE tool](https://newsletters-tool.code.dev-gutools.co.uk/) will be persisted in the CODE s3 bucket for newsletter data.

## How can we measure success?

This moves towards having the arrangement on the /email-newsletters page controlled from the newsletters tool, rather than requiring PR's in the DCR repo.

subsequent PRs:
 - [frontend repo] - read the layout data and cache from the newsletters API, include it in the data posted to DCR: https://github.com/guardian/frontend/pull/27587
 - [dcr repo] - use the layout data from frontend instead of the hard coded copy : https://github.com/guardian/dotcom-rendering/pull/12795
 - designing a user-friendly layout editor so editorial can self-serve changes
 - update which users have permissions to use the new editor

## Have we considered potential risks?

The new pages are not in the nav bar yet, so users are unlikely to find them. We can modify this branch so it won't result in the new /layouts UI pages being made live with the current config - this can be done by editing apps/newsletters-api/src/register-ui-server.ts

It should be safe to have the newsletters API publish the /layouts routes  - the data will be 'empty' unless we populate them in the UI

## Images

Screenshot 2024-09-27 at 16 49 48

Screenshot 2024-09-27 at 16 49 26

[...]

PR #331, opened 2025-01-03
layouts as objects

dblatcher:
## What does this change?

updates https://github.com/guardian/newsletters-nx/pull/328
Changes the API format for layouts to be an object with an array of groups, rather than an array of groups.

This is to make the model  easier to expand with extra data/config for the layout other than the groups of newsletters it contains.

It will also make the json data easier to handle in scala applications (EG frontend) which can more easily model the layout as (for example) a `NewsletterLayout` rather than a `List[NewsletterGroup]`.

## How to test

Should work as before locally. 
If deployed to CODE, the existing layout data will be invalidated and may need to be manually deleted from S3

new model from the api responses

https://github.com/guardian/newsletters-nx/pull/328 has not been put live yet, so there is no layout data in the PROD bucket to get invalidated by this change

### DIFF
--- a/apps/newsletters-api/static/layouts.local.json
+++ b/apps/newsletters-api/static/layouts.local.json
@@ -1,23 +1,25 @@
 {
-	"UK": [
-		{
-			"title": "Get started",
-			"newsletters": [
-				"bmx-tesla",
-				"roi-female",
-				"electric-bicycle",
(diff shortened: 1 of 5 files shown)


KaliedaRik (approved):
These changes make sense. Happy to approve

PR #332, opened 2025-01-13
layout edit UI

dblatcher:
## What does this change?

Adds a UI page route for updating the layout for a newsletter page at `/layouts/edit/{EditionId}`.

Keeps the option to edit JSON under `/layouts/edit-json/{EditionId}`.

The Layout routes are still not advertised on the main menu as the feature is not used in production yet.

## How to test

Run locally and navigate to the edit page for a layout - experiment with updating the the layout.

The editor is functional, but fairly basic and had no design input - it should serve as a prototype to demonstrate to editorial / CP . There may be further feature requests based on that.

User will be able to use the editor

## Images

Screenshot 2025-01-13 at 14 24 32

## Accessibility


apps/newsletters-ui/src/app/components/edition-layouts/LayoutEditor.tsx
```diff
@@ -0,0 +1,98 @@
...
+                        void handleSubmitUpdate(localLayout)
+                    }}>Publish update</Button>
+                {updateInProgress && <CircularProgress />}
+                {feedback === 'success' && <Alert severity="success">
+                    <Typography> Layout updated</Typography>
+                    <Typography> It will take some time for site to update</Typography>
+                </Alert>}
+                {feedback === 'failure' && <Alert severity="error">
+                    <Typography> Failed to update</Typography>
+                    <Typography> If the problem persists, please contact Central Production</Typography>
```
emdash-ie, line 59: Are these spaces at the starts of the typography elements deliberate? It looks like they're stripped out of the resulting html, and unless they have some other effect I'd probably prefer we remove them.

emdash-ie:
```suggestion
                    <Typography>Layout updated</Typography>
                    <Typography>It will take some time for site to update</Typography>
                </Alert>}
                {feedback === 'failure' && <Alert severity="error">
                    <Typography>Failed to update</Typography>
                    <Typography>If the problem persists, please contact Central Production</Typography>
```

emdash-ie: Also, out of interest: why will it take some time for the site to update?

apps/newsletters-ui/src/app/components/edition-layouts/NewsletterPicker.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+                    <ListItemButton disableGutters sx={{ paddingY: 0 }}
+                        // variant={newsletter.identityName === selectedNewsletter ? 'contained' : 'outlined'}
+                        onClick={() => setSelectedNewsletter(newsletter.identityName)}
+                    >
+                        <Checkbox
+                            checkedIcon={<RadioButtonChecked />}
+                            icon={<RadioButtonUnchecked />}
+                            size="small"
+                            // edge="start"
```
emdash-ie, line 46: Are these two comments left for a reason? Happy to keep them if they are, just checking they're not an oversight.

```suggestion
                        onClick={() => setSelectedNewsletter(newsletter.identityName)}
                    >
                        <Checkbox
                            checkedIcon={<RadioButtonChecked />}
                            icon={<RadioButtonUnchecked />}
                            size="small"
```

apps/newsletters-ui/src/app/lib/modify-layout.ts
emdash-ie, line 1: The functions in this file currently treat layouts as immutable data structures, returning new ones rather than modifying the existing ones. Is it important that they do that here?

emdash-ie (approved):
This looks good to me! It'd be good to understand why I initially couldn't get it running locally, which I believe you're currently looking into. But if it works in CODE (and I can see you tested it there) I'm happy for us to merge away and fix it later.

I left a few comments below, but none should block merging.

emdash-ie:
I [deployed this PR to CODE](https://riffraff.gutools.co.uk/deployment/view/5c5621db-bf83-4730-aa0d-2cab6e1269cd) and expected that I'd be able to see the editing UI there, but visiting https://newsletters-tool.code.dev-gutools.co.uk/layouts/edit/uk gives me a blank page with no editing UI where I expect to be able to edit [this layout](https://newsletters-tool.code.dev-gutools.co.uk/layouts/uk). Any ideas what might be going wrong there? Nothing obvious in the browser logs.

dblatcher:
> I [deployed this PR to CODE](https://riffraff.gutools.co.uk/deployment/view/5c5621db-bf83-4730-aa0d-2cab6e1269cd) and expected that I'd be able to see the editing UI there, but visiting https://newsletters-tool.code.dev-gutools.co.uk/layouts/edit/uk gives me a blank page with no editing UI where I expect to be able to edit [this layout](https://newsletters-tool.code.dev-gutools.co.uk/layouts/uk). Any ideas what might be going wrong there? Nothing obvious in the browser logs.

That'll be the user permissions - the component only displays if the user has permissions to edit layouts:
https://github.com/guardian/newsletters-nx/pull/332/files#diff-9d9d2b595a9cc064b2a59f654dab8a77ec5c6887633d84ab4c1b0a351081f2d3R31-R33

I should update this to display a suitable message rather than leaving the page blank! The links to the page won't be rendered if the user doesn't have permissions, but it woudl still be better to have the page explain why its broken.

FYI, this tool isn't connected to the [Permissions](https://permissions.code.dev-gutools.co.uk/) tool - there's a block of json listing the users with permissions in the AWS parameters store (all guardian users can see the tool, but extra permissions are required to make edits).

apps/newsletters-ui/src/app/components/edition-layouts/LayoutEditor.tsx
dblatcher: Frontend keeps a cached copy of the newsletters data from the API and only refreshes it once every 15 minutes:
https://github.com/guardian/frontend/blob/main/common/app/services/newsletters/NewsletterSignupLifecycle.scala

There's also a short cache on the Fastly config for the public version of the newsletters API (the responses are configured with a max age of 60 seconds, so shouldn't be older than that) - taken together its hard to say exactly how long it will be before the site update.

For context - there are two versions of the newsletters API:

- https://newsletters-tool.local.dev-gutools.co.uk/api/newsletters - publishes the live data and the write endpoints, but requires panda-auth to access
- https://newsletters.guardianapis.com/api/newsletters - publishes only the GET endpoint, with the caching in place

apps/newsletters-ui/src/app/lib/modify-layout.ts
dblatcher: The layouts are being used as state variables and recommendation is to treat them as readonly : https://react.dev/learn/updating-objects-in-state#treat-state-as-read-only

I suspect the components would still work as they are now if the function modified and returned the objects instead of returning new ones - but from memory you can get some odd behaviour if you use features like `memo` or `useCallback` that compare the identity of objects rather than the value.

dblatcher:
> This looks good to me! It'd be good to understand why I initially couldn't get it running locally, which I believe you're currently looking into. But if it works in CODE (and I can see you tested it there) I'm happy for us to merge away and fix it later.
> 
> I left a few comments below, but none should block merging.

Thank you - still looking into the issues running locally. To recap, I can reproduce the issues (client side build of the UI failing due to an issue in the 'popper.js' module within the MUI `styled` import) by deleting node_modules and running `npm i` on this branch, but this branch will run after running `npm i` on main and switching to this branch.

I suspect that there is something wrong with one of the import statement in the new components that is somehow causing a conflict in the client-side dependency tree (differing versions of `styled` maybe?), but haven't tracked it down as yet

apps/newsletters-ui/src/app/components/edition-layouts/LayoutEditor.tsx
emdash-ie: That makes sense, thanks for the explanation!

apps/newsletters-ui/src/app/lib/modify-layout.ts
emdash-ie: Oh yes, good point! Thanks for the explanation :)

dblatcher:
> Agreed, an error message explaining the lack of permissions would be helpful. Is it worth disabling the edit buttons with an explanation of the lack of permissions rather than removing them (to be similarly clear)? (Happy to leave both of these changes to a later PR even if we are making them!)

I personally think it would be better to render the buttons in a disabled state, so the user know that the action is possible (albeit for someone else) and doesn't go hunting around the site thinking they have missed the 'edit' button, but this point was discussed for earlier features and it was decided its better UX overall to only show the controls that are available to the user, so that's the pattern that's been followed elsewhere in the tool.

In practise, most people who go to the tool are in the newsletters team in CP, so have the 'edit' permissions anyway. If we did start showing disabled buttons, we should probably do it across the other pages too.

emdash-ie:
> I personally think it would be better to render the buttons in a disabled state, so the user know that the action is possible (albeit for someone else) and doesn't go hunting around the site thinking they have missed the 'edit' button, but this point was discussed for earlier features and it was decided its better UX overall to only show the controls that are available to the user, so that's the pattern that's been followed elsewhere in the tool.
> 
> In practise, most people who go to the tool are in the newsletters team in CP, so have the 'edit' permissions anyway. If we did start showing disabled buttons, we should probably do it across the other pages too.

Ok, a conversation for another time then – thanks for the background.

emdash-ie:
I’ve tried diffing the node_modules folder produced by `npm install` on main and on this branch and the only differences I see are in `node_modules/.cache/nx/nxdeps.json`. But none of the differences look like they should be responsible: some changes to existing files (e.g. LayoutView.tsx no longer depends on react because the import has been removed) and then new dependencies for the new files. I’ll try looking at the nx task graph and see if the changes make sense, but I don’t feel like I’m on a particularly promising path.

dblatcher:
> I’ve tried diffing the node_modules folder produced by `npm install` on main and on this branch and the only differences I see are in `node_modules/.cache/nx/nxdeps.json`. But none of the differences look like they should be responsible: some changes to existing files (e.g. LayoutView.tsx no longer depends on react because the import has been removed) and then new dependencies for the new files. I’ll try looking at the nx task graph and see if the changes make sense, but I don’t feel like I’m on a particularly promising path.

@emdash-ie , I've pushed a change that seems to fix the local build issues - I've removed the use of @emotion/styled from the components in the UI component and replaced them with the css export from '@mui/material' (which I think is essentially a re-export from @emotion/react, maybe with some mui flavour added). 

I'm not sure what it was in this PR that surfaced the problem, but I'm guessing there was some kind of underlyng conflict between the use of @emotion/styled in the generators config (nx.json) and using it in the production code.

Do you mind checking if this version works for you locally too?

PR #333, opened 2025-02-10
add layouts page links

dblatcher:
## What does this change?

Adds links to the /layouts page recently added. The pages are already in place - not that [DCR is using the data published by the API](https://github.com/guardian/dotcom-rendering/pull/12795), we should make the pages visible on the uI

## How to test

Link buttons on the header nav and home page will direct to the /layouts page

## Images

Screenshot 2025-02-10 at 11 50 47

### DIFF
--- a/apps/newsletters-ui/src/app/components/HomeMenu.tsx
+++ b/apps/newsletters-ui/src/app/components/HomeMenu.tsx
@@ -126,7 +126,7 @@ export function HomeMenu() {
 
 				<ButtonGridItem
 					path="/layouts"
-					content={'All Newsletters page layouts'}
+					content={'View Newsletter Layouts'}
 					variant="outlined"
 				/>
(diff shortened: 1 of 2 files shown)


rhystmills:
This is working as expected locally, my one thought is whether the 'content' label could be an action rather than a description, e.g. "View Newsletter Layouts" rather than "All Newsletters Page Layouts" - I can see that the other buttons on the main page are labelled with actions:

image

dblatcher:
> This is working as expected locally, my one thought is whether the 'content' label could be an action rather than a description, e.g. "View Newsletter Layouts" rather than "All Newsletters Page Layouts" - I can see that the other buttons on the main page are labelled with actions:
> 

great point - just pushing an update now

rhystmills (approved):
Running nicely locally 👍

PR #335, opened 2025-02-19
Move to bigger private subnets

JuliaBrigitte:
Frontent is moving to bigger, private subnets and an SSM Parameter /devx/bigger/private/subnets to reference them. 

## What does this change?

Set newslettersapiPrivateSubnets and newslettersapiPrivateSubnets to new bigger private subnets with parameter /devx/bigger/private/subnets

## How to test

Check if deployment still works

## Next steps

This is to be a interim step that will be reverted once we are in a position to make the values of the new values in the parameter /devx/bigger/private/subnets into a new default. So when the move to the new bigger subnet is complete.

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -29,6 +29,11 @@ deployments:
       templateStagePaths:
         CODE: NewslettersTool-CODE.template.json
         PROD: NewslettersTool-PROD.template.json
+      templateParameters:
+        newslettersapiPrivateSubnets:
+          /devx/bigger/private/subnets
+        newsletterstoolPrivateSubnets:
+          /devx/bigger/private/subnets
 
   # Newsletters internal tool autoscaling deployment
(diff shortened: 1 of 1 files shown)


riff-raff.yaml
```diff
@@ -29,6 +29,11 @@ deployments:
       templateStagePaths:
         CODE: NewslettersTool-CODE.template.json
         PROD: NewslettersTool-PROD.template.json
+      templateParameters:
```
akash1810, line 32: Is it worth adding an inline comment here to explain what's happening in case folk come across it in its temporal state?

akash1810 (approved):
LGTM!! (assuming we've tested on CODE)

riff-raff.yaml
JuliaBrigitte: Good idea, added a comment

PR #343, opened 2025-07-04
Dblatcher/switch to express

dblatcher:
## What does this change?

changes the Api  app to use Express rather than Fastify.

## Accessibility


apps/newsletters-api/src/app/routes/notifications.ts
```diff
@@ -81,11 +89,9 @@ export function registerNotificationRoutes(app: FastifyInstance) {
 		},
 	);
 
-	app.get<{
-		Params: { newsletterId: string; action: string };
-	}>(
+	app.get(
 		'/api/email/:newsletterId/:action',
-		{ preValidation: canTriggerNotification },
+		// { preValidation: canTriggerNotification },
```
waisingyiu, line 94: Can this comment be deleted?

apps/newsletters-api/src/register-ui-server.ts
```diff
@@ -1,32 +1,60 @@
...
+	Response,
+} from 'express'
+import { static as serveStatic } from 'express';
 
-export function registerUIServer(app: FastifyInstance) {
+
+const routeMap = {
+	'/': [
+		'',
+		'/templates',
```
waisingyiu, line 14: Just an enquiry - why is `/templates` added to the `routeMap` differently from other endpoints like `/launched` and `/drafts`?

In the Fastify code, it appears to have the same pattern as others?

waisingyiu (approved):
Great engineering work!  Happy to approve, with just a couple of non-blocking questions.

I could see that the CI build was successful and the app was running on CODE.

apps/newsletters-api/src/register-ui-server.ts
dblatcher: I guess I was following the structure used in the [UI application](https://github.com/guardian/newsletters-nx/blob/main/apps/newsletters-ui/src/main.tsx) to define react-router routes - the `/templates` page is a subpath of the "homeRoute" `RouteObject` rather than being on its own `RouteObject`. I think it was done this way because `/templates` is a single page with no further routing off it.

From the server point of view, it would have made more sense to have done `/templates` like `/launched` and `/drafts`, but its equivalent since the routeMap is effectively flattened into the list of paths to serve the Ui to.

It would be good to clean this up on a future PR - if we add express routers, it would be nice if the client-side routes corresponded to them.

PR #344, opened 2025-07-23
remove fastify and serverless-express

dblatcher:
## What does this change?

https://github.com/guardian/newsletters-nx/pull/343 replaced fastify with express, but neglected to remove the  fastify dependencies. It also added serverless-express in the expectation if would be needed, but it wasn't. This PR removes the redundant dependencies.

## How to test

Should build and run as before

removes the vulnerability that prompted https://github.com/guardian/newsletters-nx/pull/343

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1111,12 +1111,13 @@
 			}
 		},
 		"node_modules/@babel/code-frame": {
-			"version": "7.24.2",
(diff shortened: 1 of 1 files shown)


liamness (approved):
Can see this builds on the CI, and also verified that the project still works locally

PR #510, opened 2026-01-15
Updating qs and react router packages to avoid reported vulnerabilities

liamness:
## What does this change?

See more info in security advisories:

https://github.com/advisories/GHSA-6rw7-vpxm-498p
https://github.com/advisories/GHSA-2w69-qvjg-hvjx

Bumps `react-router-dom` to `6.30.3` and overrides `qs` to `6.14.1`. Normally wouldn't like to force a override, but the version of `body-parser` we're pulling in has it pinned to a specific version.

## How to test

Will deploy to CODE and look at testing instructions in #343

Removing dependency on package versions with reported high severity vulnerabilities.

Risk of unintentional regressions, so will test functionality, and make newsletters team aware of changes also.

### DIFF
--- a/cdk/package-lock.json
+++ b/cdk/package-lock.json
@@ -7,9 +7,6 @@
     "": {
       "name": "cdk",
       "version": "0.0.0",
-      "dependencies": {
-        "constructs": "^10.4.2"
-      },
       "devDependencies": {
         "@guardian/cdk": "^62.1.2",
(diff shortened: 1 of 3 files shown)


SHession (approved):
Looks safe enough and I've given CODE a quick look through. Worth letting the Newsletters team know as a courtesy.

PR #590, opened 2026-04-12
Phase 1 dependency upgrades

rBangay:
## What does this change?

This is the first pr in a series of pr's to try and update the core dependencies to the latest (or as close to as we can get).

In this pr we have updated:

- The nx dependecies from v15.9 to v22.6.5
- typescript from v5.5.2 to 5.9.3
- @typescript-eslint from v5 to 8.58.1
- @guardian/eslint-config-typsript from v3.0.0 to 12.0.0
- eslint from v8.15 to v8.57.1
- vite from v4 to v7 and vitest from v1 to v4

Originally the main objective of this first pr was to upgrade the nx packages, the other dependency upgrades in this pr subsequently became necessary.

The eslint upgrades resulted in a stricter rule set which meant fixing a lot of lint errors.

Where possible I have removed any overides from the package.json file. I'm hoping to remove more (if not all) of them by the end of these series of prs.


apps/newsletters-api/src/main.ts
```diff
@@ -49,6 +43,9 @@ if (isServingReadEndpoints()) {
 
 const start = async () => {
 	try {
+		if (isServingUI()) {
+			await registerUIServer(app);
+		}
```
rBangay, line 48: This moved here so that it is called once when the app starts. This was caught by the codeQL step (which happened after updating express) which noticed that the previous functionality could lead to the function being called multiple times, and it didn't have any rate limit protection around it

georgerichmond (approved):
Great - just one tiny comment doesn't really matter though

PR #598, opened 2026-04-16
dependency upgrades (phase 5) - typescript

### LINKED ISSUE #597: Newsletters-nx core dependency upgrades - phase 5
As a follow on from recent work to upgrade some transitive dependencies in order to fix vulnerabilities.

We had to fix the vulnerability issues using the overrides section of the package.json file as there wasn't a clear path to upgrade the core dependencies which in some cases are multiple major versions behind the latest version.

This card represents the work to upgrade our core dependencies to the latest versions.

Included in phase 5 is (hopefully the last of these for a while):

Typescript

rBangay:
## What does this change?

The fifth pr in a series of pr's in order to try and get our core dependencies up to date. This time we have updated Typescript from v5 to v6

### DIFF
--- a/apps/newsletters-ui/src/vite-env.d.ts
+++ b/apps/newsletters-ui/src/vite-env.d.ts
@@ -1,0 +1,1 @@
+/// <reference types="vite/client" />
--- a/libs/state-machine/project.json
+++ b/libs/state-machine/project.json
@@ -6,14 +6,18 @@
 	"tags": [],
 	"targets": {
 		"build": {
-			"executor": "@nx/js:tsc",
-			"outputs": ["{options.outputPath}"],
+			"executor": "nx:run-commands",
+			"outputs": ["{workspaceRoot}/dist/libs/state-machine"],
+			"inputs": ["production", "^production"],
+			"cache": true,
+			"dependsOn": ["^build"],
 			"options": {
-				"outputPath": "dist/libs/state-machine",
-				"main": "libs/state-machine/src/index.ts",
-				"tsConfig": "libs/state-machine/tsconfig.lib.json",
-				"assets": ["libs/state-machine/*.md"],
-				"rootDir": "."
+				"commands": [
+					"npx tsc -p libs/state-machine/tsconfig.lib.json",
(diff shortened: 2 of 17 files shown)


apps/newsletters-ui/src/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
rBangay, line 1: I had to add this in so that `./apps/newsletters-ui/src/app/components/GeneratedCodeDataPoint.tsx` can continue to import a css file!

libs/state-machine/project.json
```diff
@@ -6,14 +6,18 @@
...
+			"dependsOn": ["^build"],
 			"options": {
-				"outputPath": "dist/libs/state-machine",
-				"main": "libs/state-machine/src/index.ts",
-				"tsConfig": "libs/state-machine/tsconfig.lib.json",
-				"assets": ["libs/state-machine/*.md"],
-				"rootDir": "."
+				"commands": [
+					"npx tsc -p libs/state-machine/tsconfig.lib.json",
+					"cp libs/state-machine/*.md dist/libs/state-machine/ 2>/dev/null || true"
```
rBangay, line 17: I had to add these manual tsc and copy commands in here instead of using nx/js:tsc. I do not know why nx didn't want to run tsc properly after the typescript update 😖

libs/state-machine/tsconfig.lib.json
```diff
@@ -1,10 +1,18 @@
 {
 	"extends": "./tsconfig.json",
 	"compilerOptions": {
-		"outDir": "../../dist/out-tsc",
+		"outDir": "../../dist/libs/state-machine",
```
rBangay, line 4: this has changed here as part of the switch to using the 2 manual `tsc` and `cp` commands in the [‎libs/state-machine/project.json](https://github.com/guardian/newsletters-nx/pull/598/changes#diff-d035f53feb8308a27ab66fb665fa45ad4b0dc7e629a1e4e095b070f5dfbc6117) file. This was previously handled by the `@nx/js:tsc` step but now the outDir needs to explicitly match the directory that the `cp` command in project.json uses.

georgerichmond (approved):

PR #600, opened 2026-04-17
Implement a simple client-side feature switch functionality

andrewHEguardian:
## What does this change?

Adds client side feature switch functionality that can toggle switches on or off by URL and is stored in local storage.

This change is made with a view for the redesign using the `Stand` design system

Tested locally by going to https://newsletters-tool.local.dev-gutools.co.uk/?use-stand=false and https://newsletters-tool.local.dev-gutools.co.uk/?use-stand=true and seeing an extra line of HTML render or not.

---
  - https://app.asana.com/0/0/1214074623090663

### DIFF
--- a/apps/newsletters-ui/src/app/featureSwitches.ts
+++ b/apps/newsletters-ui/src/app/featureSwitches.ts
@@ -1,11 +1,12 @@
 export const featureSwitches = {
-	'use-stand': false,
-} as const;
+	'switch-stand': false,
+} as const satisfies Record<`switch-${string}`, boolean>;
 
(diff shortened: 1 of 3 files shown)


apps/newsletters-ui/src/app/featureSwitches.ts
```diff
@@ -0,0 +1,22 @@
+export const featureSwitches = {
+	'use-stand': false,
```
coldlink: wondering if all switches should be prefixed with a specific value so we don't ever accidentially clash with other local storage/query param values?

e.g. make sure every switch has a `switch-` prefix or something?

georgerichmond (approved):

PR #607, opened 2026-04-21
Fix 2 new zod data types not recognised elsewhere in codebase

rBangay:
## What does this change?

The introduction of the 2 new zod types ZodURL and ZodEmail as part of the series of dependency upgrade prs, specifically (https://github.com/guardian/newsletters-nx/pull/596) caused issues elsewhere in the codebase where the values were being compared to a ZodString type. 

Where values were previously being typed as a ZodString, they are now typed as ZodURL or ZodEmail where applicable.

This pr/fix updates the rest of the codebase to ensure wherever we check for the ZodString type we also include the ZodURL and ZodEmail types as well ... some of these cases were picked up in this follow up pr (https://github.com/guardian/newsletters-nx/pull/606) but not all of them :(

### DIFF
--- a/apps/newsletters-ui/src/app/components/SchemaForm/SchemaField.tsx
+++ b/apps/newsletters-ui/src/app/components/SchemaForm/SchemaField.tsx
@@ -255,7 +255,11 @@ export function SchemaField<T extends z.ZodRawShape>({
 	}
 
 	if (innerZod instanceof ZodArray) {
-		if (innerZod.element instanceof ZodString) {
+		if (
+			innerZod.element instanceof ZodString ||
+			innerZod.element instanceof ZodURL ||
+			innerZod.element instanceof ZodEmail
+		) {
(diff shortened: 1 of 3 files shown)


apps/newsletters-ui/src/app/components/SchemaForm/SchemaField.tsx
```diff
@@ -255,7 +255,11 @@ export function SchemaField<T extends z.ZodRawShape>({
 	}
 
 	if (innerZod instanceof ZodArray) {
-		if (innerZod.element instanceof ZodString) {
+		if (
+			innerZod.element instanceof ZodString ||
+			innerZod.element instanceof ZodURL ||
+			innerZod.element instanceof ZodEmail
+		) {
```
rBangay, line 262: It might be worth creating a new type that encompass these, but for now I wanted to get a quick fix out

georgerichmond (approved):

PR #612, opened 2026-04-21
Create redesign Newsletter creation flow with Stand form elements

andrewHEguardian:
## What does this change?

Adds a new newsletter creation flow (based on new designs). The flow is a duplicate of the existing one with form fields rearranged, and some renaming of labels. (This will be iterated upon and refined further). It also duplicates the form rendering components but with [Stand](https://github.com/guardian/stand) components.

The new flow is accessible behind the feature switch `switch-stand`.

Tested locally, there should be no change visible to users.

## Images

Different step layout
image

Stand form fields rendered

image

---
  - https://app.asana.com/0/0/1214074623398091


coldlink (approved):
Looks like a good starting point for now.

Since it's behind a feature switch we can merge in and fix/iterate any issues that might be encountered during full implementation

PR #616, opened 2026-04-22
add a new drop down input field that reverts to a text input field if…

### LINKED ISSUE #588: Add newsletter frequency dropdown field alongside free text entry
## Problem
Currently, newsletter frequency is captured as a free text field in the system. This allows flexibility but introduces inconsistency and makes downstream processing or UI consistency more difficult.

## Proposal
- Add a new drop-down field for newsletter frequency.
- The existing free text field should be kept for now.
- When a value is selected in the drop-down, that value should be served. If not, fall back to serving the original free text field value.
- The options for the drop-down will be finalized later.

## Notes
- Please structure the code so that migrating away from the free text field in the future will be straightforward.
- Ensure backwards compatibility so existing data entry forms and any backend usage continue to function.
- Add tests to validate the new field's integration.

cc @georgerichmond

rBangay:
## What does this change?

Adds a dropdown select input instead of a free text input field for the newsletter frequency input. Comes with some default options to select from and a custom option that if selected shows a text input field.

The new field should appear on the create and edit newsletters pages.

## Images
Before             |  After            |  After (custom)
:-------------------------:|:-------------------------:|:-------------------------:
  |    |  

The frequency instructions have been updated to reflect these changes too:
Screenshot 2026-04-22 at 15 41 28

### DIFF
--- a/apps/newsletters-ui/src/app/components/SchemaForm/FrequencySelectInput.tsx
+++ b/apps/newsletters-ui/src/app/components/SchemaForm/FrequencySelectInput.tsx
@@ -1,0 +1,121 @@
+import type { SelectChangeEvent } from '@mui/material';
+import {
+	FormControl,
+	FormHelperText,
+	InputLabel,
+	MenuItem,
+	Select,
+	TextField,
+} from '@mui/material';
+import type { FormEventHandler, FunctionComponent } from 'react';
+import { useEffect, useState } from 'react';
+import type { FieldProps } from './util';
+
+const FREQUENCY_OPTIONS = [
+	'Daily',
+	'Weekdays',
+	'Twice weekly',
+	'Weekly',
+	'Fortnightly',
+	'Monthly',
+] as const;
+
+const CUSTOM_KEY = 'custom';
+
+const frequencyOptionSet = new Set<string>(FREQUENCY_OPTIONS);
+
+const isExistingCustomValue = (value: string | undefined): value is string =>
+	value !== undefined && value !== '' && !frequencyOptionSet.has(value);
+
+export const FrequencySelectInput: FunctionComponent<
+	FieldProps & {
+		value?: string;
+		inputHandler: (value?: string) => void;
+	}
+> = ({
+	value,
+	optional,
+	inputHandler,
+	label = 'Frequency',
+	readOnly,
+	error,
+}) => {
+	const [isInCustomMode, setIsInCustomMode] = useState<boolean>(() =>
+		isExistingCustomValue(value),
+	);
+	const [localCustomValue, setLocalCustomValue] = useState<string>(() =>
+		isExistingCustomValue(value) ? value : '',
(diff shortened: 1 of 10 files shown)


apps/newsletters-ui/src/app/components/SchemaForm/FrequencySelectInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+				))}
+				<MenuItem value={CUSTOM_KEY}>Custom</MenuItem>
+			</Select>
+			{!isInCustomMode && error && <FormHelperText>{error}</FormHelperText>}
+			{isInCustomMode && (
+				<TextField
+					fullWidth
+					label="Custom frequency"
+					value={localCustomValue}
+					onInput={handleTextInput}
```
mgosz-guardian, line 112: Is there a reason to use `onInput` instead of `onChange` here?

apps/newsletters-ui/src/app/components/SchemaForm/util.ts
```diff
@@ -62,7 +70,7 @@ function fieldValueIsRightType(value: FieldValue, field: FieldDef): boolean {
 	const innerZod = recursiveUnwrap(field.zod);
 
 	if (innerZod instanceof ZodEnum) {
-		if (field.zod.isOptional() && typeof value === 'undefined') {
+		if (field.zod instanceof ZodOptional && typeof value === 'undefined') {
```
mgosz-guardian, line 73: Should we update the SchemaField file to use `ZodOptional` too for consistency? Seems that "'zod.isOptional' is deprecated"

mgosz-guardian (approved):

apps/newsletters-ui/src/app/components/SchemaForm/util.ts
rBangay: yeah good spot 👍

apps/newsletters-ui/src/app/components/SchemaForm/FrequencySelectInput.tsx
rBangay: Hey, it should work either way, just a habit for input fields. Apparently `onInput` fires as soon as the value of the input has changed as opposed to `onChange` that fires once the value has changed and the element has lost focus. `onInput` will also not fire if you change the value via js.

*Edit according to mdn the input event doesn't always fire if the input is a checkbox or a radio input if the user toggles it (https://developer.mozilla.org/en-US/docs/Web/API/Element/input_event) despite the fact it should (https://html.spec.whatwg.org/multipage/input.html#the-input-element:event-input-2). This wouldn't affect us here as the input type is a text field. I think I will leave this here, but make a note not to use `onInput` in the future for checkboxes or radio buttons. `onInput` is used elsewhere in the codebase where we are dealing with textArea's

PR #620, opened 2026-04-24
Use Stand Top bar component

andrewHEguardian:
## What does this change?

This implements the Stand top bar component as part of the redesign, behind a feature switch. Users should not see any changes unless they add the url param `?switch-stand=true`.

Tested locally

## Images
Before
image

After
image


andrewHEguardian:
Closing because I've made a mess with rebasing...

PR #630, opened 2026-05-08
Stand step nav wide breakpoints

oliverabrahams:
## What does this change?

Updates the Stand redesign flow to use a new stepper side nav component. This is part of the redesign seen here in [figma](https://www.figma.com/design/gmHnKr3tOzvnYp1FZeIl4t/Stand--Editorial-tool-Design-System--WIP%F0%9F%9A%A7-?node-id=446-10264&t=FipizOsPE5qgFMg0-0). 

- a button as the pages of the wizard do not have different urls.
- `<ol> <li> </li> </ol>` for some additional semantic meaning.
- aria-current="step" on the active button is the right signal.

TODO in another PR
- [ ]  mobile breakpoints

It has only been tested locally but this is all behind the feature switch so shouldn't affect any of the core flow for now.

Before this PR there is a bug in the stand flow where it breaks on step 2. This needs to be looked at separately.

Low risk as behind a feature switch.

## Images

Screenshot 2026-05-08 at 13 06 32
Screenshot 2026-05-08 at 13 06 14
Screenshot 2026-05-08 at 13 06 01

## Accessibility

---
  - https://app.asana.com/0/0/1214229634760779

### DIFF
--- a/apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
+++ b/apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
@@ -3,6 +3,7 @@ import { baseColors, semanticColors } from '@guardian/stand';
 import { Icon } from '@guardian/stand/icon';
 import { Typography } from '@guardian/stand/typography'
 import { CheckCircleOutlined, CircleSharp, WarningAmberOutlined } from '@mui/icons-material';
+import type { ReactNode} from 'react';
 import { useState } from 'react';
 import { Button } from 'react-aria-components';
 import { resolveStepStatus, StepStatus } from '@newsletters-nx/state-machine';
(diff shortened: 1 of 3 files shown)


apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
```diff
@@ -1,11 +1,15 @@
...
-	StepButton,
-	StepLabel,
-	Stepper,
-	Typography,
-} from '@mui/material';
+	CheckCircleOutlined,
+	CircleSharp,
+	WarningAmberOutlined,
+} from '@mui/icons-material';
+import type { ReactNode } from 'react';
```
coldlink: Rather than using the svg icons from mui, we already have the material symbols font installed which is our recommended icon set in stand, so you can just pass the name of the icon into the stand `Icon` component!

e.g.

```
<Icon ...>check_circle_outline</Icon>
```

While the names are typed within the Icon component itself, it might be useful within stand to also export the [`MaterialSymbol`](https://github.com/guardian/stand/blob/mm/layout-changes/src/fonts/material-symbols-types.ts#L3856) type somewhere.

I guess you could also do `type Symbol = IconProps['symbol']` which would achieve the same thing

apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
```diff
@@ -21,63 +25,117 @@ interface Props {
...
+	stepNumber: number;
+	onClick: () => void;
+	ariaLabel: string;
+	isDisabled: boolean;
+	description: string;
+};
+
+type StatusRowProps = {
+	label: string;
+	icon: ReactNode;
```
coldlink: e.g.

```suggestion
	icon: Pick<IconProps['symbol'], 'check_circle_outline' | 'circle' | 'warning_amber';
```

apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
```diff
@@ -21,63 +25,117 @@ interface Props {
...
-					Optional{' '}
-					<span role="img" aria-label="green-cross">
-						❎
-					</span>
-				</Typography>
+				<StatusRow
+					label="Optional"
+					isDisabled={isDisabled}
+					iconColor={semanticColors.text['success-inverse']}
+					icon={<CircleSharp />}
```
coldlink: so..

```suggestion
					icon="circle"
```

apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
```diff
@@ -21,63 +25,117 @@ interface Props {
...
-					Complete{' '}
-					<span role="img" aria-label="checkmark">
-						✅
-					</span>
-				</Typography>
+				<StatusRow
+					label="Complete"
+					isDisabled={isDisabled}
+					iconColor={semanticColors.text.success}
+					icon={<CheckCircleOutlined />}
```
coldlink: ...

```suggestion
					icon="check_circle_outline"
```

apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
```diff
@@ -21,63 +25,117 @@ interface Props {
...
-					Incomplete{' '}
-					<span role="img" aria-label="cross">
-						❌
-					</span>
-				</Typography>
+				<StatusRow
+					label="Incomplete"
+					isDisabled={isDisabled}
+					iconColor={semanticColors.text.error}
+					icon={<WarningAmberOutlined />}
```
coldlink:
```suggestion
					icon="warning_amber"
```

apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
```diff
@@ -152,58 +210,166 @@ export const StandRedesignStepNav = ({
...
+		}
+	`;
+	return (
+		// Use a button instead of a link here because it is a single page application
+		<Button
+			css={buttonStyles}
+			isDisabled={isDisabled}
+			aria-label={ariaLabel}
+			aria-current={isCurrent ? 'step' : undefined}
+			onClick={() => onClick()}
```
coldlink:
```suggestion
		// Use a button instead of a link here because it is a single page application
		<Button
			css={buttonStyles}
			isDisabled={isDisabled}
			aria-label={ariaLabel}
			aria-current={isCurrent ? 'step' : undefined}
			onPress={() => onClick()}
```

React Aria prefers [`onPress`](https://react-aria.adobe.com/Button#events)

apps/newsletters-ui/src/app/components/StandRedesignStepNav.tsx
```diff
@@ -21,63 +25,117 @@ interface Props {
...
-	if (active) {
-		return `${description} (current step)`;
+	const statusDescription =
+		status === StepStatus.Complete
+			? 'complete'
+			: status === StepStatus.Incomplete
+				? 'incomplete'
+				: status === StepStatus.Optional
+					? 'optional'
+					: undefined;
```
coldlink: maybe this would be slightly easier to read as an iife? Something like this?


```suggestion
	const statusDescription = (() => {
		switch (status) {
			case StepStatus.Complete:
				return 'complete';

			case StepStatus.Incomplete:
				return 'incomplete';

			case StepStatus.Optional:
				return 'optional';

			default:
				return undefined;
		}
	})();
```

oliverabrahams: I agree much better like this

coldlink (approved):

PR #647, opened 2026-05-19
add Vite 8/Rolldown browser stubs for @aws-sdk

### LINKED ISSUE #646: fix newsletters-nx dev run bug
The following dependency update pr (https://github.com/guardian/newsletters-nx/pull/643) introduced a bug when trying to run newsletters-nx locally via ```npm run dev```

Specifically the updates of the nx packages from version 22.6.5 to 22.7.2. The nx packages have a transitive dependency on vite which was bumped up to version 8 with the 22.7.2 version bump.

Vite version 8 changed the bundler it uses from esbuild to rolldown. 

Rolldown (Vite 8's new bundler) does strict static analysis of re-export chains (which we have in libs/newsletters-data-client/src/index.ts and libs/util/src/index.ts) and throws an error when it encounters node:fs imports during dep optimisation (server imports in the client). This was silently ignored by esbuild (Vite ≤=7).

A reasonable fix for now (until eithger Rolldown can handle the server imports in the client bundle more gracefully, or we refactor the way we import files across the codebase) is to create some stubs in the vite config file (the file that is used when you run the site locally) for the offending server imports that are finding their way through to the client bundle that return null.

rBangay:
## Background
The following dependency update pr (https://github.com/guardian/newsletters-nx/pull/643) introduced a bug when trying to run newsletters-nx locally via `npm run dev` failed.

Specifically the updates to the nx packages from version 22.6.5 to 22.7.2. The nx packages have a transitive dependency on vite which was bumped up to version 8 with the 22.7.2 version bump.

Vite version 8 changed the bundler it uses from esbuild to rolldown.

Rolldown (Vite 8's new bundler) does strict static analysis of re-export chains (which we have in libs/newsletters-data-client/src/index.ts and libs/util/src/index.ts) and throws an error when it encounters node:fs imports during dep optimisation (server imports in the client). This was silently ignored by esbuild (Vite ≤=7).

## What does this change?

Adds two plugins to Vite in vite.config.ts:
- awsSdkBrowserStub: runs during the dev server module graph (when the browser requests a file). Transforms @aws-sdk value imports in source files to inline stub classes so that the browser never requests any @aws-sdk paths
- awsSdkOptimizerFix: Intercepts @aws-sdk/credential-provider-web-identity and node:fs inside the optimiser before Rolldown hits the broken re-export chain. Without this, the optimiser processes @aws-sdk/credential-providers and in doing so tries to follow `export * from "./fromTokenFile"` which imports `node:fs` which Rolldown can't analyse in browser mode and crashes.

Also adds `sideEffects: false` to newsletters-data-client to improve
tree-shaking in production builds.

### DIFF
--- a/apps/newsletters-ui/vite.config.ts
+++ b/apps/newsletters-ui/vite.config.ts
@@ -1,9 +1,78 @@
 /* eslint-disable -- We want default export for config files */
 /// <reference types="vitest" />
-import { defineConfig } from 'vite';
+import { defineConfig, type Plugin } from 'vite';
 import react from '@vitejs/plugin-react';
 import { nxViteTsPaths } from '@nx/vite/plugins/nx-tsconfig-paths.plugin';
 import path from 'path';
+
+// @aws-sdk packages use Node.js built-ins (fs, crypto, etc.) unavailable in the
+// browser. The UI never calls AWS APIs directly — all AWS calls go via the API
+// server. This transform hook rewrites @aws-sdk value imports in project source
+// files to inline stub classes before Vite ever resolves or pre-bundles them,
+// avoiding both "missing export" and "Outdated Optimize Dep" (504) errors.
+const awsSdkBrowserStub = (): Plugin => ({
+	name: 'aws-sdk-browser-stub',
+	enforce: 'pre',
+	transform(code, id) {
+		if (id.includes('/node_modules/')) return null;
+		if (!code.includes('@aws-sdk/')) return null;
+
+		const result = code
+			// Named value imports — but NOT `import type { ... }`
+			.replace(
+				/^import\s+(?!type[\s{])(\{[^}]+\})\s+from\s+['"]@aws-sdk\/[^'"]+['"]\s*;?/gm,
+				(_, imports) =>
+					imports
+						.slice(1, -1)
+						.split(',')
+						.map((s: string) => s.trim())
+						.filter(Boolean)
+						.map((imp: string) => {
+							const local = imp.split(/\s+as\s+/).pop()!.trim();
+							return `const ${local} = class ${local} {};`;
+						})
+						.join('\n'),
+			)
+			// Namespace imports: import * as foo from '@aws-sdk/...'
+			.replace(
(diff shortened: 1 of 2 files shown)


mgosz-guardian (changes requested):
Looks like Vite upgrade exposed a pre-existing mixed browser/server export boundary in newsletters-data-client. The stubs are a workaround. I think we should try to fix the root problem. Could we try to split browser-safe and server-only entrypoints and migrate UI imports to browser-only, then remove stubs?

rBangay:
> Looks like Vite upgrade exposed a pre-existing mixed browser/server export boundary in newsletters-data-client. The stubs are a workaround. I think we should try to fix the root problem. Could we try to split browser-safe and server-only entrypoints and migrate UI imports to browser-only, then remove stubs?

Hey, yep you're right. I think we should. I think that work might be a fair bit bigger so I thought I'd get this fix out first. We'd have to refactor the code base to properly separate out the server/shared code from the client code.

georgerichmond (approved):
Approved - I did a clean checkout and it works - I think we should investigate moving to pnpm so that we can separate out the package.json for packages for the front and back and as they don't really intersect. Also moving to pnpm aligns with dotcom-rendering and is a better developer experience.

rBangay:
> Approved - I did a clean checkout and it works - I think we should investigate moving to pnpm so that we can separate out the package.json for packages for the front and back and as they don't really intersect. Also moving to pnpm aligns with dotcom-rendering and is a better developer experience.

Agreed, I've created a card to look at seperating out the server and client code properly. I'll start work on that shortly, which will get rid of the changes in this pr

mgosz-guardian (approved):

mgosz-guardian:
> Looks like Vite upgrade exposed a pre-existing mixed browser/server export boundary in newsletters-data-client. The stubs are a workaround. I think we should try to fix the root problem. Could we try to split browser-safe and server-only entrypoints and migrate UI imports to browser-only, then remove stubs?

Approved on a basis that it is a temp solution and proper fix will follow.

PR #659, opened 2026-06-01
[Stand Redesign] Add custom/other optional input to radio

coldlink:
## What does this change?

Updates the radio input group to allow for a custom/other value for scenarios where the provided options are not sufficient

This is done by updating the `StandRadioSelectInput` which wraps the `@guardian/stand` `Radio` and `RadioGroup` to include a prop which enables the custom text input field when the option is selected.

Much of the work is down to managing the state of the component regarding the value and change of the value.

Applies it to the create newsletters flow `frequency` field:

https://github.com/user-attachments/assets/f1c00958-241b-4472-9629-1047f4e50e78

---
  - https://app.asana.com/0/0/1215329238947031

### DIFF
--- a/apps/newsletters-ui/src/app/components/StandRedesignSchemaForm/RadioSelectInput.tsx
+++ b/apps/newsletters-ui/src/app/components/StandRedesignSchemaForm/RadioSelectInput.tsx
@@ -1,40 +1,114 @@
+import { baseSpacing } from '@guardian/stand';
 import { Radio, RadioGroup } from '@guardian/stand/RadioGroup';
+import { TextInput } from '@guardian/stand/TextInput';
+import type { TextInputProps } from '@guardian/stand/TextInput';
+import { useEffect, useState } from 'react';
+import type { FormEventHandler } from 'react';
 import type { FunctionComponent } from 'react';
 import type { StandardFormProps } from './SchemaField';
 import type { FieldProps } from './util';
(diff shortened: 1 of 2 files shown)


apps/newsletters-ui/src/app/components/StandRedesignSchemaForm/RadioSelectInput.tsx
```diff
@@ -1,40 +1,119 @@
...
+
+	useEffect(() => {
+		if (hasCustomValue) {
+			setIsOtherSelected(true);
+			return;
+		}
+
+		if (typeof value === 'string' && options.includes(value)) {
+			setIsOtherSelected(false);
+		}
```
oliverabrahams: If someone is typing in other and types one of the options it will switch back to that option but they may not have finished typing their custom string

oliverabrahams (commented):
This looks great one minor UX comment

apps/newsletters-ui/src/app/components/StandRedesignSchemaForm/RadioSelectInput.tsx
coldlink: Good spot! Have just removed this if block, as it doesn't actually matter if someone types in the same thing that already exists as another option,

oliverabrahams (approved):
Looks great

PR #670, opened 2026-06-08
[Stand Redesign] Add 1px right border to grid

coldlink:
## What does this change?

Adds the 1px right border to all the pages in the stand redesign wizard between the main form and the right side bar, only for `lg` breakpoint.

To accommodate this we removed all top/bottom spacing from the Layout, made the height of the grid `100%`, and instead applied the spacing to individual items instead to get the spacing correct. This could be backported to stand.

The border itself was applied by using an empty `<Item>` and applying the css on that due to the alignment of where the border was required according to the figma designs.

<table>
<tr>
<th> lg
<th> md
<th> sm
<tr>
<td>
Screenshot 2026-06-08 at 14-42-31 Newsletters tool
<td>
Screenshot 2026-06-08 at 14-42-50 Newsletters tool
<td>
Screenshot 2026-06-08 at 14-43-08 Newsletters tool
</table>

---
  - https://app.asana.com/0/0/1215511324467940

### DIFF
--- a/apps/newsletters-ui/src/app/components/StandRedesignWizard.tsx
+++ b/apps/newsletters-ui/src/app/components/StandRedesignWizard.tsx
@@ -1,7 +1,8 @@
 import { css } from '@emotion/react';
-import { baseSpacing, semanticColors } from '@guardian/stand';
+import { baseSpacing, semanticColors, semanticGrid } from '@guardian/stand';
 import { Grid, Item } from '@guardian/stand/Grid';
 import { Layout } from '@guardian/stand/Layout';
+import { from } from '@guardian/stand/utils';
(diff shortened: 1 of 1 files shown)


oliverabrahams (approved):
Looks good, What was happening with the SkipConfirmationDialog before?

coldlink:
> Looks good, What was happening with the SkipConfirmationDialog before?

The skipConfirmationDialog was taking up a 0px height full 12 cols width below the form regardless of it it visible or not. So I just adjusted the width on desktop to get the border item correct.

Not sure how to exactly trigger this dialog though?

it might be another form? e.g might be used in the edit form or something?
as the wizard is shared between all similar form interfaces

we'll see when we come across it 😓

PR #672, opened 2026-06-08
Fix | Stand Redesign |  Restore keyboard focus for RadioGroup when no value is selected

coldlink:
## What does this change?

When a field's initial value is an empty string "", react-aria's useRadio treated it as a controlled selection ("" != null), found no matching radio, and left every radio with tabIndex=-1, making the group unreachable by keyboard. Passing undefined instead of "" lets react-aria fall into its uncontrolled path and correctly assign tabIndex=0 to the first radio.

<table>
<tr>
<th> Before
<th> After
<tr>
<td>

https://github.com/user-attachments/assets/ca204130-c84a-40eb-a6c6-05c2beedc7e5

<td>

https://github.com/user-attachments/assets/9197da5f-c839-4553-95fe-e104ecf9ce01

</table>

Fixes: https://github.com/guardian/stand/issues/127

---
  - https://app.asana.com/0/0/1215513748232166


oliverabrahams (approved):
Nicee

PR #687, opened 2026-07-02
[Stand Redesign] Apply design review changes where possible

coldlink:
## What does this change?

- Top bar navigation text -> Only capitalise first letter
- Update text and spacing
- See [Figma](https://www.figma.com/design/jD7ZXQ2hmna9qQjUBu3ioH/Newsletter-pilot?node-id=3095-23655&t=Puld6MeEQsQB52ga-1) for full list of changes
  - The only change not made was making the H1 subtext into body-lg, as this didn't quite easily work in markdown, and it's not applied on every page

<table>
<tr>
<th> Page
<th> Before
<th> After
<tr>
<td> Create Newsletters - Introduction
<td>
Screenshot 2026-07-02 at 08-32-29 Newsletters tool
<td>
Screenshot 2026-07-02 at 08-32-23 Newsletters tool
<tr>
<td> Create Newsletters - Name and frequency
<td>
Screenshot 2026-07-02 at 08-33-42 Newsletters tool
<td>
Screenshot 2026-07-02 at 08-33-38 Newsletters tool
<tr>
<td> Create Newsletters - Tags
<td>
Screenshot 2026-07-02 at 08-36-29 Newsletters tool
<td>
Screenshot 2026-07-02 at 08-36-33 Newsletters tool
<tr>
<td> Article Rendering - Headings/Text
<td>
Screenshot 2026-07-02 at 08-38-43 Newsletters tool
<td>
Screenshot 2026-07-02 at 08-38-47 Newsletters tool
<tr>
<td> Launch Wizard - Headings/Text
<td>
Screenshot 2026-07-02 at 08-39-36 Newsletters tool
<td>
Screenshot 2026-07-02 at 08-39-41 Newsletters tool
</table>

---
  - https://app.asana.com/0/0/1216165837061490

### DIFF
--- a/apps/newsletters-ui/src/app/components/StandRedesignMarkdownView.tsx
+++ b/apps/newsletters-ui/src/app/components/StandRedesignMarkdownView.tsx
@@ -1,5 +1,9 @@
 import { css } from '@emotion/react';
-import { semanticSpacing, semanticTypography } from '@guardian/stand';
+import {
+	semanticColors,
+	semanticSpacing,
+	semanticTypography,
+} from '@guardian/stand';
 import type { IconProps } from '@guardian/stand/Icon';
 import { Icon } from '@guardian/stand/Icon';
 import { Link } from '@guardian/stand/Link';
@@ -44,7 +48,12 @@ const remarkLinkDirective: Plugin = () => (tree) => {
 interface MarkdownViewProps {
 	markdown: string;
 	bottomSpacing?: keyof typeof semanticSpacing;
-	componentTypographyOverrides?: Partial<Record<'H1' | 'H2' | 'H3', TypographyProps['variant']>>;
+	componentTypographyOverrides?: Partial<
+		Record<
+			'H1' | 'H2' | 'H3' | 'P' | 'STRONG' | 'LI' | 'LI_STRONG',
+			TypographyProps['variant']
+		>
+	>;
 }
 
 const isExternal = (href?: string) => !href?.startsWith('/');
@@ -64,7 +73,10 @@ const LinkWithNewTabIfExternal = (props: {
 	);
 };
 
-const H1 = (props: {typographyVariant?: TypographyProps['variant']; children?: ReactNode}) => {
+const H1 = (props: {
+	typographyVariant?: TypographyProps['variant'];
+	children?: ReactNode;
+}) => {
 	return (
 		<Typography
 			element="h1"
@@ -80,7 +92,10 @@ const H1 = (props: {typographyVariant?: TypographyProps['variant']; children?: R
 	);
 };
 
-const H2 = (props: {typographyVariant?: TypographyProps['variant']; children?: ReactNode}) => {
+const H2 = (props: {
+	typographyVariant?: TypographyProps['variant'];
+	children?: ReactNode;
+}) => {
 	return (
 		<Typography
 			element="h2"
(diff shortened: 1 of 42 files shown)


apps/newsletters-ui/src/app/components/StandRedesignMarkdownView.tsx
```diff
@@ -140,11 +163,39 @@ const UlMarginOverride = (props: { children?: ReactNode }) => {
...
+
+const TypographyStrong = (props: {
+	typographyVariant?: TypographyProps['variant'];
+	children?: ReactNode;
+}) => {
 	return (
 		<Typography
-			element="span"
-			variant="bodyBoldSm"
+			element="b"
```
jamesmockett, line 197: Is there any reason this isn't a `strong` element as the component name suggests?

apps/newsletters-ui/src/app/components/StandRedesignWizard.tsx
```diff
@@ -341,6 +339,7 @@ export const StandRedesignWizard: React.FC<WizardProps> = ({
 								display: flex;
 								flex-direction: row;
 								gap: ${semanticSpacing.stackMd};
+								margin-top: ${semanticSpacing.stackXl};
```
jamesmockett, line 342: Does the `FailureAlert` container above still need a bottom margin applied to it now this container has a top margin? Based on my (somewhat hazy 😅) understanding of margin collapsing the two margins will collapse and the larger one here will win.

apps/newsletters-ui/src/app/components/StandRedesignSchemaForm/index.tsx
```diff
@@ -73,13 +73,13 @@ export function StandRedesignSchemaForm<T extends z.ZodRawShape>({
 		});
 	}
 
-	return (
+	return fields.length > 0 ? (
 		<div
 			css={css`
 				display: flex;
 				flex-direction: column;
 				gap: ${baseSpacing['32Rem']};
```
jamesmockett: Should this also use semantic rather than base spacing? eg. `semanticSpacing.stackLg`

apps/newsletters-ui/src/app/components/StandRedesignMarkdownView.tsx
coldlink: Screenshot 2026-07-02 at 16 06 52

Our typography component doesn't currently like `strong` as a valid element especially as it's a semantic element, we should probably investigate/change this. Will add a card into a backlog. `b` will suffice for now i think, it's better than using `span`!

apps/newsletters-ui/src/app/components/StandRedesignWizard.tsx
coldlink: I'll play around with the `FailureAlert` and see what happens.

coldlink: Your understanding is right! We don't need the margin bottom anymore, but we do need a margin-top now!

Screenshot 2026-07-02 at 16-12-58 Newsletters tool

coldlink: Will update the margin-bottom to be margin-top. When we get updated designs for the alert container we can fully correct the spacing we need.

apps/newsletters-ui/src/app/components/StandRedesignMarkdownView.tsx
jamesmockett: Ah, okay. Yeah, `b` is definitely preferable to a `span` in that case 👍

apps/newsletters-ui/src/app/components/StandRedesignWizard.tsx
jamesmockett: > We don't need the margin bottom anymore, but we do need a margin-top now!

Ha! I suppose having the margins all on one side avoids any confusion about margin collapsing at least.

RikRootsGuardian (approved):
I've run the branch locally and can confirm that requested changes have been implemented. Example screenshots below:

Image
Image
Image
Image

